PR #8, opened 2023-02-12
Basic implementation of agent

fmarek-kindred:


packages/talos_agent/src/agent.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    /// Starts listener task which reads Talos decisions from the topic
+    pub fn start(&self, int_type: &TalosIntegrationType) -> Result<(), String> {
+        let agent_id = self.config.agent_id.clone();
+        let config = self.kafka_config.clone().expect("Kafka configuration is required");
+        let in_flight = Arc::clone(&self.in_flight);
+
+        let it = int_type.clone();
+        tokio::spawn(async move {
+            let consumer: Box<ConsumerType> = match it {
```
gk-kindred, line 66: Would be nice in future iterations to be able to pass the Consumer based on `trait` and therefore I can pass the actual consumer or any mock to run my tests. 

Currently I can pass only the real consumer or a mock which will return `None` on receive message. 

Not needed in this PR, but would be great to have this refactored later.

packages/talos_agent/src/api.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    pub fn build(&self) -> Result<Box<TalosAgentType>, String> {
+        let publisher: Box<PublisherType> = match self.integration_type {
+            Kafka => {
+                let config = &self.kafka_config.clone().expect("Kafka configuration is required");
+                let kafka_publisher = KafkaPublisher::new(self.config.agent_id.clone(), config);
+                Box::new(kafka_publisher)
+            }
+            _ => Box::new(MockPublisher {}),
```
gk-kindred, line 111: Same comment as the `consumer`. A particular implementation for Mock is tightly coupled here. We should make it flexible to pass different Mocks to test different scenarios.

packages/talos_agent/src/main.rs
```diff
@@ -0,0 +1,193 @@
...
+        if d < min {
+            min = d;
+        }
+        if d > max {
+            max = d;
+        }
+        debug!("Transaction has been certified. Details: {:?}", rsp);
+    }
+
+    metrics.sort();
```
gk-kindred, line 134: nice one with the metrics :)

gk-kindred (approved):

packages/talos_agent/src/api.rs
fmarek-kindred: I agree, currently it is just a placeholder really. I will change this part.

PR #9, opened 2023-02-14
feat: suffix pruning

gk-kindred:


packages/talos_suffix/src/utils.rs
```diff
@@ -0,0 +1,8 @@
+/// Retrives all the items that are valid and not None.
+pub fn get_nonempty_suffix_items<'a, T: 'a>(items: impl Iterator<Item = &'a Option<T>> + 'a) -> impl Iterator<Item = &'a T> + 'a {
```
anil0906, line 2: Test can be written for these without the context of suffix

packages/talos_suffix/src/suffix.rs
```diff
@@ -18,11 +21,29 @@ impl<T> Suffix<T>
...
 {
-    pub fn new(capacity: usize) -> Suffix<T> {
-        let suffix_vec: Vec<Option<SuffixItem<T>>> = vec![None; capacity + 1];
-        let messages = VecDeque::from(suffix_vec);
-
-        let meta = SuffixMeta { head: 0, prune_vers: None };
+    pub fn with_config(config: SuffixConfig) -> Suffix<T> {
+        let SuffixConfig {
+            capacity,
+            prune_start_threshold,
```
anil0906, line 27: I think we can add docs on core components. Like meaning and use case of new variables. It becomes easier to understand the change.

anil0906 (approved):

packages/talos_suffix/src/utils.rs
gk-kindred: updated in latest commit

packages/talos_suffix/src/suffix.rs
gk-kindred: Basic documentation added just for this method for now. Will add more documentation in subsequent PRs

PR #10, opened 2023-02-17
Feature/actor model

fmarek-kindred:
This is draft PR because I need to rebase this work onto more recent master. There are some merging conflicts.


packages/talos_agent/src/agentv2/agent_v2.rs
```diff
@@ -0,0 +1,58 @@
...
+        agent_config: AgentConfig,
+        kafka_config: Option<KafkaConfig>,
+        int_type: &TalosIntegrationType,
+        certify: (Sender<CertifyRequestChannelMessage>, Receiver<CertifyRequestChannelMessage>),
+        cancel: (Sender<CancelRequestChannelMessage>, Receiver<CancelRequestChannelMessage>),
+        publish_times: &Arc<Mutex<HashMap<String, u64>>>,
+    ) -> Result<TalosAgentImplV2, String> {
+        let state_manager = StateManager::new(agent_config, kafka_config, int_type, publish_times).unwrap();
+
+        state_manager.start(certify.1, cancel.1).await;
```
gk-kindred, line 27: This shouldn't be done inside the `new` function. Ideally we only expect a new instance of `Self/Struct` to be created here.

examples/agent_client/examples/agent_client.rs
```diff
@@ -0,0 +1,324 @@
...
+
+async fn certify(batch_size: i32) -> Result<(), String> {
+    info!("Certifying {} transactions", batch_size);
+
+    // "global" state where we will collect publishing times: A - B, where
+    // B: transaction start time
+    // A: the end of call to kafka send_message(candidate)
+    let publish_times: Arc<Mutex<HashMap<String, u64>>> = Arc::new(Mutex::new(HashMap::new()));
+    let mut tasks = Vec::<JoinHandle<Result<Timing, String>>>::new();
+    let agent = Arc::new(make_agentv2(&publish_times).await);
```
gk-kindred, line 109: Just use `Arc::clone(publish_times)` rather than passing the reference of the clone all the way down, and using clone in the `new` method of `StateManager`

examples/agent_client/examples/agent_client.rs
```diff
@@ -0,0 +1,324 @@
...
+        let ac = Arc::clone(&agent);
+        let task = tokio::spawn(async move { Timing::capture(|| async { ac.certify(make_candidate(Uuid::new_v4().to_string())).await }).await });
+        tasks.push(task);
+
+        let p = i + 1;
+        if p % PROGRESS_EVERY == 0 || p == batch_size {
+            info!("Published {} of {}", p, batch_size);
+        }
+
+        if done % adjust_every == 0 {
```
gk-kindred, line 138: Little confused on this one, shouldn't the delay be done where the task is consumed and awaited? 

Because when you introduce the delay here, it just delays the rate at which the task is added to the Vec, but not the rate at which it is executed.

i.e the ` for task in tasks ` loop?

scripts/consolidate-logs_test-concurrent-agents.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
gk-kindred, line 1: Could you remove the `agent-log-1.txt` and `agent-log-2.txt` files? Guess it was created as part of some script you ran?

packages/talos_agent/src/agentv2/agent_v2.rs
```diff
@@ -0,0 +1,58 @@
...
+#[async_trait]
+impl TalosAgent for TalosAgentImplV2 {
+    async fn certify(&self, request: CertificationRequest) -> Result<CertificationResponse, String> {
+        let (tx, mut rx) = mpsc::channel::<CertificationResponse>(1);
+
+        let m = CertifyRequestChannelMessage::new(&request, &tx);
+
+        let to_state_manager = self.rx_certify.clone();
+        let response = match to_state_manager.send(m).await {
+            Ok(()) => match rx.recv().await {
```
gk-kindred, line 45: I understand why you do this here from performance testing perspective. But in real world scenario this should be avoided. There is no guarantee that we are receiving the response of the message we send. It could be some other message. 

Why you don't notice this problem in performance testing is because you run a for loop of tasks which are run serially.

examples/agent_client/examples/agent_client.rs
fmarek-kindred: Yep that is the intention. Delay before spawning a new task so that our intended rate is matching real rate. No need to wait for task to finish before spawning a new one. Otherwise publishing will become sequential, right?

scripts/consolidate-logs_test-concurrent-agents.sh
fmarek-kindred: Good catch. Thanks.

packages/talos_agent/src/agentv2/agent_v2.rs
fmarek-kindred: Hold on. This channel instance is for the specific request. The only message we will receive through it is the reply to our request, no other reply can come to this channel instance.

packages/talos_agent/src/agentv2/agent_v2.rs
```diff
@@ -0,0 +1,58 @@
...
+        certify: (Sender<CertifyRequestChannelMessage>, Receiver<CertifyRequestChannelMessage>),
+        cancel: (Sender<CancelRequestChannelMessage>, Receiver<CancelRequestChannelMessage>),
+        publish_times: &Arc<Mutex<HashMap<String, u64>>>,
+    ) -> Result<TalosAgentImplV2, String> {
+        let state_manager = StateManager::new(agent_config, kafka_config, int_type, publish_times).unwrap();
+
+        state_manager.start(certify.1, cancel.1).await;
+
+        Ok(TalosAgentImplV2 {
+            rx_certify: certify.0,
```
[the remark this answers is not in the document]
fmarek-kindred: Done. Thanks :)

scripts/consolidate-logs_test-concurrent-agents.sh
fmarek-kindred: Done

examples/agent_client/examples/agent_client.rs
gk-kindred: when you spawn a task and attach it to a handle, the task is executed only when you  `await` on the handle. i.e in the above code, the publish actually happens inside the for loop of `tasks`. Therefore it is actually sequential at the moment.

packages/talos_agent/src/agentv2/agent_v2.rs
gk-kindred: This is related to [this comment ](https://github.com/kindredgroup/talos/pull/10#discussion_r1116552094) I just made. your perf test is currently running 1 task at a time. And that is why this is fine. Ideally when agent publishes in parallel, there is no guarantee that the certifier will return the decisions in the same order the candidate messages were received.

examples/agent_client/examples/agent_client.rs
gk-kindred: Sorry you are right, I just got mixed up with something else. Ignore this comment :)

packages/talos_agent/src/agentv2/agent_v2.rs
fmarek-kindred: Thanks. This is fixed now.

examples/agent_client/examples/agent_client.rs
fmarek-kindred: Thanks. This is fixed now.

packages/talos_agent/src/agentv2/agent_v2.rs
gk-kindred: `mpsc` can have only 1 receiver. Since you create only one instance of `agent` and then put `agent` in `Arc` it's basically the same receiver being used across threads. Could you check and confirm on this?

fmarek-kindred: Are we talking about lines 43 (variable "rx") and line 49 (variable "rx"). Its new every time someone calls certify()

gk-kindred: ignore this comment as well, I was looking at the wrong `receiver` all along :(

gk-kindred (approved):

PR #13, opened 2023-02-28
feat: Use assign and seek instead of subscribe and sleep.

fmarek-kindred:


packages/talos_agent/src/messaging/kafka.rs
```diff
@@ -137,49 +135,47 @@ impl KafkaConsumer {
...
+        // This line is required for seek operation to be successful.
+        partition_list.set_partition_offset(topic, partition, Offset::End).unwrap();
+
+        let rslt_assign = self.consumer.assign(&partition_list);
+        match rslt_assign {
+            Ok(()) => {
+                return match self.get_offset(partition, Duration::from_secs(5)) {
+                    Ok(offset) => {
+                        log::info!("Seeking on partition {} to offset: {:?}", partition, offset);
+                        match self.consumer.seek(topic, partition, offset, Duration::from_secs(5)) {
```
gk-kindred, line 152: Too many nested `match` statements can make it hard to read. 

And example to simplify this would be:-

```rust
self.consumer
                            .seek(topic, partition, offset, Duration::from_secs(5))
                            .map_err(|e| format!("Error when Seeking to offset {:?}. Error: {}", offset, e))
```

packages/talos_agent/src/agentv2/state_manager.rs
```diff
@@ -61,13 +62,22 @@ impl StateManager {
 
         let it_for_consumer = self.int_type.clone();
         let config_for_consumer = config.clone();
+
+        let consumer_barrier = Arc::new(Notify::new());
+        let consumer_barrier_arc = Arc::clone(&consumer_barrier);
         tokio::spawn(async move {
```
gk-kindred, line 68: Maybe in future, it would be good to move this thread logic as a separate service all together. It listens to the decisions coming in, and passes it to the core logic of  the `state_manager.rs`, which is to receive messages and manage the state.

packages/talos_agent/src/messaging/kafka.rs
```diff
@@ -137,49 +135,47 @@ impl KafkaConsumer {
...
 
-                Ok(())
-            }
-            Err(kafka_error) => {
-                error!("Error when subscribing to topics. {:?}", kafka_error);
-                Err(kafka_error.to_string())
+    fn get_offset(&self, partition: i32, timeout: Duration) -> Result<Offset, String> {
+        let now_ms = OffsetDateTime::now_utc().unix_timestamp() * 1000;
+        let topic = self.config.certification_topic.as_str();
+        match self.consumer.offsets_for_timestamp(now_ms, timeout) {
```
gk-kindred, line 168: Although this works, just wondering do we just use the `highest watermark` as the broker can easily provide the high and low watermark on partition rather than searching for the offsets for a particular time, across all partitions (although we are single partition).

And then also you have the overhead of searching the partition and topic to get the offset. 

vs just using this:-

https://docs.rs/rdkafka/latest/rdkafka/consumer/trait.Consumer.html#tymethod.fetch_watermarks

fmarek-kindred: I will try to use watermarks.

packages/talos_agent/src/messaging/kafka.rs
gk-kindred: Flattening the entire code would be as follows:- 

```rust
rslt_assign.map_err(|e| format!("Error invoking assign('{}', {}. Error: {})", topic, partition, e))?;

let offset = self
    .get_offset(partition, Duration::from_secs(5))
    .map_err(|e| format!("Error fetching offset of partition {}. Error: {})", partition, e))?;

Ok(self
    .consumer
    .seek(topic, partition, offset, Duration::from_secs(5))
    .map_err(|e| format!("Error when Seeking to offset {:?}. Error: {}", offset, e))?)
```

gk-kindred: No nested match statements. But I was able to do this because of the  `?` operator. You can read more about it here https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html

fmarek-kindred: I like it better. Thanks!

fmarek-kindred: Done

packages/talos_agent/src/messaging/kafka.rs
fmarek-kindred: Done

gk-kindred (approved):

packages/talos_agent/src/agentv2/state_manager.rs
fmarek-kindred: Done

gk-kindred (approved):

PR #16, opened 2023-02-28
Feature/env based adapters

anil0906:


packages/talos_certifier_adapters/src/kafka/config.rs
```diff
@@ -1,49 +1,57 @@
...
     pub brokers: Vec<String>,
-    pub topic_prefix: String,
-    pub consumer_topic: String,
-    pub producer_topic: String,
+    pub topic: String,
     pub client_id: String,
     pub group_id: String,
     pub username: String,
     pub password: String,
+    pub producer_config_overrides: HashMap<&'a str, &'a str>,
```
gk-kindred, line 13: Why not use `String` for key and value, rather than doing the above? Because now the compiler isn't sure of the lifetime of those key/value and we are forced to explicitly set the lifetime. 

But at the same time, if you look at `talos_certifier_adapters/src/certifier_kafka_pg.rs`, we are explicitly setting the ` 'a ` on the root function. i.e we are saying we expect these to live through the entire life of the root function. Therefore why not just use `Hashmap<&'static str, &'static str>`?

anil0906: In this case we might as well use static lifetimes. But as a general practice i think we should try to use more reference rather than owned types.

ekoutanov: Or a `Cow` with a static lifetime. I'd have no beef with that approach.

gk-kindred (approved):

PR #17, opened 2023-03-03
refactor: Init kafka publisher and consumer

fmarek-kindred:


packages/talos_agent/src/messaging/kafka.rs
```diff
@@ -43,6 +45,36 @@ impl KafkaPublisher {
         setup_kafka_auth(&mut cfg, kafka);
         cfg.create().expect("Unable to create kafka producer")
     }
+
+    async fn send_init_message(&self, header_key: String, id: String, timeout_ms: u64) -> Result<(), String> {
```
fmarek-kindred, line 49: This will crash Talos until it filters corrupted messages.

PR #27, opened 2023-03-14
feature: Agent upgrade

fmarek-kindred:
- Improved handling of errors, use typed errors
- Improved directory layout
- Encapsulated metrics tracker into agent as a feature
- Removed old agent.

Rebased on the most recent master.


packages/talos_agent/src/agent/errors.rs
```diff
@@ -0,0 +1,55 @@
...
+            kind: CertificationTimout { xid, elapsed_ms },
+            reason: "Timeout".to_string(),
+            cause: None,
+        }
+    }
+}
+
+impl From<MessagingError> for AgentError {
+    fn from(e: MessagingError) -> Self {
+        AgentError {
```
gk-kindred, line 37: The `MessageErrorKind` granularity is lost if you want to behave differently for different error kinds.

packages/talos_agent/src/agent/errors.rs
```diff
@@ -0,0 +1,55 @@
...
+    Messaging,
+    Internal,
+}
+
+#[derive(Debug, ThisError)]
+#[error("Talos agent error: '{kind}'.\nReason: {reason}\nCause: {cause:?}")]
+pub struct AgentError {
+    pub kind: AgentErrorKind,
+    pub reason: String,
+    pub cause: Option<Box<dyn Error + Send>>,
```
gk-kindred, line 22: Not sure about the advantage of using `Box<dyn Error + Send>` in the cause vs just using `String` and mapping the internal error as `e.to_string()`. As all errors implement `Display` and usually `e.to_string()` will have sufficient information. 

Could you check the advantage of boxing and sending the full error here?

Also in rust library coding `Box<dyn Error>` is generally discouraged as the error is ambiguous to the consumer of the library

fmarek-kindred: I have changed it to be Optional<String>. But commit will be in the next branch where I am covering code with tests.

packages/talos_agent/src/agent/errors.rs
fmarek-kindred: Let's discuss it tomorrow. Not sure how much of low level details client will want to know. Kafka has lots of error codes, I guess we need to pick few critical ones.

gk-kindred (approved):

PR #28, opened 2023-03-21
fix: Improving test coverage

fmarek-kindred:


gk-kindred (approved):
Too many generics because we are trying to mock the channels as well. But apart from that the PR looks good

PR #35, opened 2023-04-17
feat: initial commit for replicator

gk-kindred:
Base implementation:-

- Listens to consumer message and writes to suffix.
- Listens to decision message and updates the safepoint and decision for the candidate message in suffix.
- At set interval picks a batch of messges.
  - Filters only the messages that has safepoint >= snapshot
  - Filters only the messages that has a statemap


packages/cohort/src/replicator/core.rs
```diff
@@ -0,0 +1,67 @@
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use talos_certifier::model::DecisionMessage;
```
fmarek-kindred, line 5: @gk-kindred we might need to reduce copies of classes which represent the same things like decision message, candidate message. 

We already have two copies - one inside certifier, the other is inside agent. Now [the third is coming in](https://github.com/kindredgroup/talos/pull/35/files#diff-b80d6ec230a253a3fe5b04ec3ec5229d9e207c94004e59768d13767c34c10bdeR45). Given that cohort is already including agent as dependency (which comes with its data model) I am wondering does it make sense to additionally include certifier data model dependency?

packages/cohort/src/replicator/kafak_consumer.rs
fmarek-kindred, line 1: typo in the file name "kafk"

packages/cohort/src/replicator/replicator_service.rs
```diff
@@ -0,0 +1,127 @@
...
+impl<S, M> Replicator<S, M>
+where
+    S: talos_suffix::SuffixTrait<CandidateMessage> + SuffixDecisionTrait<CandidateMessage> + Debug,
+    M: MessageReciever<Message = ReceiverMessage> + Send + Sync,
+{
+    pub fn new(receiver: M, suffix: S) -> Self {
+        Replicator { receiver, suffix }
+    }
+
+    pub async fn run(&mut self) {
```
fmarek-kindred, line 31: Think about splitting this into smaller methods for easier maintenance and unit testing later on.

packages/cohort/src/replicator/core.rs
gk-kindred: Yeah this was temporary as I needed `statemap` field which `agent` and `certifier` version of candidate message didn't have yet. Was bit unsure if I would break anything so was holding off till rest of the functionality was added in

packages/cohort/src/replicator/kafak_consumer.rs
gk-kindred: Changed locally, will push this soon.

packages/cohort/src/replicator/core.rs
```diff
@@ -0,0 +1,128 @@
...
+            suffix,
+            _phantom: PhantomData,
+        }
+    }
+
+    pub(crate) async fn process_consumer_message(&mut self, version: u64, message: T) {
+        if version > 0 {
+            self.suffix.insert(version, message).unwrap();
+        } else {
+            warn!("Version 0 will not be inserted into suffix.")
```
fmarek-kindred, line 95: If you have access to XID here, could be useful to print it.

fmarek-kindred (approved):

PR #36, opened 2023-04-19
refactor: Delegate bank and snapshot operations to DB layer

fmarek-kindred:


packages/cohort/src/main.rs
```diff
@@ -1,22 +1,29 @@
...
     tokio::spawn(async move {
-        let agent = Cohort::make_agent(cfg_agent, cfg_kafka).await;
-        let mut cohort = Cohort::new(agent);
+        let agent = Cohort::init_agent(cfg_agent, cfg_kafka).await;
 
-        cohort.start().await;
+        let database = Database::init_db(cfg_db.clone()).await;
+        let cohort = Cohort::new(agent, Arc::clone(&database));
+
+        prefill_db(database).await;
```
gk-kindred, line 23: Can we seed the database with initial data as a separate binary? This was if we want to test `replicator` or rerun `cohort initiator` without resetting the db everytime.

packages/cohort/src/bank_api.rs
```diff
@@ -0,0 +1,209 @@
...
+
+#[async_trait]
+impl Action for AccountUpdate {
+    async fn execute<T>(&self, client: &T) -> Result<(), String>
+    where
+        T: GenericClient + Sync,
+    {
+        let params: &[&(dyn ToSql + Sync)] = &[&self.amount, &(self.new_version as i64), &self.account.number];
+
+        let statement = client.prepare(Self::sql(self.update_version)).await.unwrap();
```
gk-kindred, line 128: Could you use `prepare_cached` here?

packages/cohort/src/state/postgres/database.rs
```diff
@@ -0,0 +1,93 @@
...
+        result.iter().map(fn_converter).collect::<Vec<T>>()
+    }
+
+    pub async fn batch<T>(&self, batch: Vec<T>) -> Result<(), String>
+    where
+        T: Action,
+    {
+        let mut client = self.pool.get().await.unwrap();
+        let tx = client.transaction().await.unwrap();
+        for action in batch.iter() {
```
gk-kindred, line 78: It will be nice to try pipelining the queries in batch to improve performance

https://docs.rs/tokio-postgres/0.4.0-rc.3/tokio_postgres/#pipelining

packages/cohort/src/main.rs
fmarek-kindred: This action is idempotent, can run multiple times without side effect as long as file contains the lower version number. than in DB.

We can think of separating pre-fill into another script.

packages/cohort/src/state/postgres/database.rs
fmarek-kindred: Thanks. This is TODO item. I will make a JIRA ticket.

gk-kindred (approved):

PR #38, opened 2023-05-01
feat: Initiator checks for the safe point before installing

fmarek-kindred:


packages/cohort/src/core.rs
```diff
@@ -288,8 +288,16 @@ impl Cohort {
 
         // Talos gave "go ahead"
 
+        // Check safepoint condition before installing ...
+
+        let safepoint = resp.safepoint.unwrap(); // this is safe for 'Committed'
```
gk-kindred, line 293: This will `panic` when the decision is aborted as `safepoint` will be `None`

packages/cohort/src/core.rs
```diff
@@ -288,8 +288,16 @@ impl Cohort {
...
+        SnapshotApi::await_until_safe(Arc::clone(&self.database), safepoint).await?;
+
+        // install
+
         op_impl(Arc::clone(&self.database), AccountRef::new(account.number.clone(), None), amount, resp.version).await?;
 
+        // TODO: this should be done by replicator
         SnapshotApi::update(Arc::clone(&self.database), resp.version).await?;
```
gk-kindred, line 301: Initiator should not update `snapshot` as initiator does out of order install.

packages/cohort/src/core.rs
```diff
@@ -332,17 +340,23 @@ impl Cohort {
             return Ok(());
         }
 
-        // Talos gave "go ahead"
+        // Check safepoint condition before installing ...
+
+        let safepoint = resp.safepoint.unwrap(); // this is safe for 'Committed'
```
gk-kindred, line 345: Same as previous comment regarding `panic`. 

Also we shouldn't have this logic here in the `transfer` method, because replicator calls the same method and this waiting doesn't apply for replicator.

packages/cohort/src/core.rs
gk-kindred: You already have a comment in your code, so ignore this :-)

packages/cohort/src/core.rs
fmarek-kindred: There is a "return" on line 286.

packages/cohort/src/core.rs
fmarek-kindred: ok

packages/cohort/src/core.rs
fmarek-kindred: RE: panic. There is return statement on line 340

This is example usage coming from "generate_workload".  I don't think replicator calls this one. 
It calls database.batch  for now, but with [new PR](https://github.com/kindredgroup/talos/pull/41) it will call BatchExecutor::execute(Vec<StatemapItem>)

gk-kindred (approved):

PR #40, opened 2023-05-02
feat: replicator integration to cohort - first iteration

gk-kindred:
* Commit to db and kafka are commented out.
* Pruning of suffix is not implemented.

### DIFF
--- a/packages/cohort/src/bin/replicator.rs
+++ b/packages/cohort/src/bin/replicator.rs
@@ -89,7 +89,7 @@ async fn main() {
     //  c. Create suffix.
     let suffix_config = SuffixConfig {
         capacity: 10,
-        prune_start_threshold: None,
+        prune_start_threshold: Some(2),
         min_size_after_prune: None,
     };
     let suffix: Suffix<ReplicatorCandidate> = Suffix::with_config(suffix_config);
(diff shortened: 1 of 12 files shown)


packages/cohort/src/bin/replicator.rs
```diff
@@ -1,15 +1,76 @@
...
+        if sm_item.action == "bank_transfer" {
+            let new_version = sm_item.version;
+            let payload = serde_json::from_value::<ReplicatorBankTransfer>(sm_item.payload).unwrap();
+
+            acc.push(Box::new(Transfer::new(
+                AccountRef::new(payload.from, Some(new_version)),
+                AccountRef::new(payload.to, Some(new_version)),
+                payload.amount.to_string(),
+                new_version,
+                true,
```
fmarek-kindred, line 46: this should be `true` only for the last item in the list because SQL statements are written something like that

UPDATE x=..., version=v WHERE version < v; 
UPDATE x=..., version=v WHERE version < v;  // this will not execute skipped if line above updated 'v'.

fmarek-kindred: But you dont worry about it, in the new batch api I am handling this internally.

fmarek-kindred (approved):

PR #42, opened 2023-05-02
feat: Allow batch of statemap items to include unknown business action

fmarek-kindred:


packages/cohort/src/tx_batch_executor.rs
```diff
@@ -24,97 +24,90 @@ impl BatchExecutor {
...
-                                tx_error.unwrap_err(),
-                            ))
-                        } else {
-                            Err(format!("Cannot update snapshot. Error: {}. Rollback", snapshot_error))
-                        };
-                    }
-                }
+        if affected_rows == 0 {
+            // still fail here
+            return Err(Self::handle_rollback(tx.rollback().await, "No rows where updated".to_string()));
```
gk-kindred, line 48: I feel this shouldn't throw an error, but an early return with 0 rows. And maybe print a warning?

gk-kindred: Because sometimes, replicator has already moved ahead and if initiator is already installed it, we wont install it again. 

And vice-versa.

PR #44, opened 2023-05-04
feat: Optimase internal structures

fmarek-kindred:
- Starting to make internal structure more organised:
- introduced new packages: "actions", "model/requests" and moved classes accordingly.
- moved data model classes into model/ package to keep then under one roof
- removed updated version_flag from action making them trust that version can only move forward, that simplified SQLs, so now, we always update item versions.


gk-kindred (approved):
One thing I would probably do different is co-locate all the bank specific items in a  `banks` folder.

``` 
banks/
   api.rs
   actions/
      // bank specific actions
   models/
      // bank specific models
```

PR #46, opened 2023-05-11
feat: Replicator updates snapshot on aborts

gk-kindred:
* Fixes the bug on suffix
* Updates the snapshot even on aborts

### DIFF
--- a/packages/cohort/src/replicator/pg_replicator_installer.rs
+++ b/packages/cohort/src/replicator/pg_replicator_installer.rs
@@ -1,0 +1,28 @@
+// $coverage:ignore-start
+use std::io::Error;
+
+use async_trait::async_trait;
+use log::info;
+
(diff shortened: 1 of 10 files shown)


packages/cohort/src/replicator/pg_replicator_installer.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub struct PgReplicatorStatemapInstaller {
+    pub pg: PostgresApi,
+}
+
+#[async_trait]
+impl ReplicatorInstaller for PgReplicatorStatemapInstaller {
+    async fn install(&mut self, sm: Vec<StatemapItem>, version: Option<u64>) -> Result<bool, Error> {
+        info!("Last version ... {:#?} ", version);
+        info!("Original statemaps received ... {:#?} ", sm);
```
fmarek-kindred, line 19: these two should be debug. When we run high load and measure perf, that info will show up a lot

fmarek-kindred (approved):

PR #51, opened 2023-06-26
Load generator at controlled rate.

fmarek-kindred:
Added new package "examples_support" to keep code needed in for examples. Cargo layout does not allow to create additional modules used only for examples. Currently it contains only one class extracted from debugging branch. The helper to generate items at specified rate. Useful for performance tests.

### DIFF
--- a/packages/talos_agent/src/messaging/kafka.rs
+++ b/packages/talos_agent/src/messaging/kafka.rs
@@ -54,7 +54,9 @@ impl KafkaPublisher {
         cfg.set("bootstrap.servers", &kafka.brokers)
             .set("message.timeout.ms", &kafka.message_timeout_ms.to_string())
             .set("queue.buffering.max.messages", "1000000")
-            .set("topic.metadata.refresh.interval.ms", "4")
+            .set("topic.metadata.refresh.interval.ms", "5000")
+            .set("socket.keepalive.enable", "true")
+            .set("acks", "0")
             .set_log_level(kafka.log_level);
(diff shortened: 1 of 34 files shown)


packages/talos_agent/src/messaging/kafka.rs
```diff
@@ -54,7 +54,9 @@ impl KafkaPublisher {
         cfg.set("bootstrap.servers", &kafka.brokers)
             .set("message.timeout.ms", &kafka.message_timeout_ms.to_string())
             .set("queue.buffering.max.messages", "1000000")
-            .set("topic.metadata.refresh.interval.ms", "4")
+            .set("topic.metadata.refresh.interval.ms", "5000")
+            .set("socket.keepalive.enable", "true")
+            .set("acks", "0")
```
gk-kindred, line 59: We probably shouldn't put `acks=0` for real runs?

ekoutanov: I don't believe this is a problem.
The implication of `acks=0` is that a message may appear to be published to the client (i.e., agent) whereas it hasn't been stably recorded on the broker(s), possibly losing the message as a result.
The effect is that the decision will never come, for lack of candidate, and the agent will retry as of the eventual timeout.

fmarek-kindred: Has this branch passed the review?

gk-kindred (approved):

PR #52, opened 2023-06-26
Add utilities to pre-fill cohort DB

fmarek-kindred:


scripts/cohort/data-generator.py
gk-kindred, line 1: Although this is fine to use python. Would have preferred a script or rust, so that we don't have to worry about maintaining one more language while refactoring

fmarek-kindred: I usually use  python as "shell" script, because the latter is none sense with it comes to anything involving variables :)

gk-kindred: Just from a maintainability with respect to other devs and maintenance at deploy if we use it in any environment. 

Would be nice to avoid using another language

gk-kindred (approved):

PR #54, opened 2023-06-27
feat: Cohort with built-in replicator

fmarek-kindred:


packages/cohort/src/replicator2/service.rs
```diff
@@ -0,0 +1,167 @@
...
+                    }
+                }
+
+                opt_install_resp = rx_install_response.recv() => {
+                    if let Some(InstallOutcome::Success { version, started_at: _started_at, finished_at: _finished_at }) = opt_install_resp {
+                        // ack += 1;
+                        if let Err(e) = replicator.update_suffix(version) {
+                            log::warn!("Error updating suffix for version: {}. Error: {}", version, e);
+                        } else {
+                           // let _ = replicator.receiver.commit(version).await;
```
gk-kindred, line 160: We will need to look at the `commit` strategy. One way to do would be similar to that done on `certifier` to maintain `tpl (topic partition list)` and at set interval commit using `tpl`

packages/cohort/src/replicator2/cohort_suffix.rs
```diff
@@ -0,0 +1,368 @@
...
+            0
+        };
+
+        self.messages
+            .iter()
+            .skip(skip_count)
+            .flatten()
+            .take_while(|i| i.is_decided)
+            .filter(|i| !i.is_installed && !i.is_in_flight)
+            .take(1)
```
gk-kindred, line 256: This could be reason why the performance is lower when I compared the performance between the way `replicator2` creates statemap vs the changes I made in this [branch](https://github.com/kindredgroup/talos/blob/feature/replictor-perf-updates/packages/cohort/src/replicator/suffix.rs#L97).

We do all the iteration, flatten and filter and take only 1 out of it. And this function is called in a loop.

In the branch I shared above, you can see, we pick all the eligible records, and later there is a loop to further narrow down  and create the statemap.

packages/cohort/src/replicator2/cohort_suffix.rs
```diff
@@ -0,0 +1,368 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct SuffixItem {
+    /// Data received from Kafka Consumer
+    pub item: ReplicatorCandidate2,
+    pub item_ver: u64,
+    pub decision_ver: Option<u64>,
+    /// Flag to denote if this item has been decided (is_decided is set to true irrespective of the outcome).
+    pub is_decided: bool,
+    pub is_in_flight: bool,
```
gk-kindred, line 19: As we install serially, only the latest is important to us. Do we need `is_in_flight`? 

The `latest_in_flight` prop on the `CohortReplicator` should be enough for filtering the ones we don't need.

gk-kindred:
We have lot of metric code in each functions, which at times makes it hard to focus on the actual functionality the function is doing. 

- Would be nice if we can move them at later point into `macros` so the code looks cleaner. 
- Also the ability to turn on and off as required.
- Also we should move the metrics logging from `warn` to `trace` or `debug` maybe?

packages/cohort/src/replicator2/service.rs
fmarek-kindred: https://jira.kindredgroup.com/browse/KDT-67

packages/cohort/src/replicator2/cohort_suffix.rs
fmarek-kindred: GK is raising JIRA ticket for this item based on new temp branch

packages/cohort/src/replicator2/cohort_suffix.rs
fmarek-kindred: This items is in progress in GK branch linked above.

fmarek-kindred:
> We have lot of metric code in each functions, which at times makes it hard to focus on the actual functionality the function is doing.
> 
> * Would be nice if we can move them at later point into `macros` so the code looks cleaner.
> * Also the ability to turn on and off as required.
> * Also we should move the metrics logging from `warn` to `trace` or `debug` maybe?

https://jira.kindredgroup.com/browse/KDT-65

gk-kindred (approved):

PR #55, opened 2023-06-28
feat: Pack Talos decision metrics into structure. Add binary to print metrics using kafka message as source.

fmarek-kindred:
1. As noted in previous PR, introducing metrics structure in Talos DecisionMessage.
2. Use that data to print CSV histogram (new  binary created)

### DIFF
--- a/packages/cohort/src/metrics.rs
+++ b/packages/cohort/src/metrics.rs
@@ -1,5 +1,7 @@
 use std::{collections::HashMap, time::Duration};
 
+use metrics::model::MinMax;
+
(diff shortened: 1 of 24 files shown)


packages/cohort/src/metrics.rs
```diff
@@ -36,56 +38,6 @@ impl Default for TxExecSpans {
     }
 }
 
-#[derive(Debug, Clone)]
-pub struct MinMax {
```
fmarek-kindred, line 40: Moved to packages/metrics

gk-kindred (approved):

PR #56, opened 2023-06-30
refactor: Change bank_account DB schema to store attributes as separate columns

fmarek-kindred:
- Change DB schema
- Add flag to suppress metrics printing
- Remove unused deposit and withdraw actions. Keep only transfer.


packages/cohort/src/model/json/bank_account_deserialize.rs
gk-kindred, line 1: This custom deserialialize file is not really needed anymore. Just use the `serder(Serailize,Deserialize)` against the struct

packages/cohort/src/model/json/bank_account_serialize.rs
gk-kindred, line 1: As mentioned for deserialize

packages/cohort/src/bin/sql/V1__create_tables.sql
```diff
@@ -1,10 +1,11 @@
 CREATE TABLE IF NOT EXISTS bank_accounts (
-    "number"    VARCHAR(255) PRIMARY KEY,
-    "data"      JSONB NOT NULL
+    "number"    VARCHAR(20) PRIMARY KEY,
+    "name"      VARCHAR(20) NOT NULL,
+    "amount"    VARCHAR(20) NOT NULL,
```
gk-kindred, line 4: why use `VARCHAR` for amount instead of `DECIMAL`

packages/cohort/src/state/postgres/data_store.rs
```diff
@@ -78,7 +83,16 @@ impl DataStore {
     }
 
     pub fn account_from_row(row: &Row) -> BankAccount {
-        row.get::<&str, Json<BankAccount>>("data").0
+        BankAccount {
+            name: row.get::<&str, String>("name"),
+            number: row.get::<&str, String>("number"),
+            version: row.get::<&str, i64>("version") as u64,
+            balance: Money::from_str(
+                row.get::<&str, String>("amount").as_str(),
```
gk-kindred, line 91: `rusty_money` package we use stores the amount as `Decimal`. 

image

And if we use `bank_accounts.balance` as `DECIMAL` instead of `VARCHAR`, we could use `Money::from_decimal` instead of `Money::from_str. Which is way  simpler and less over head. 

Code of `Money::from_str`
image


vs Code from `Money::from_decimal`
image

gk-kindred (commented):
Nice one! Have few comments, please have a look

packages/cohort/src/state/postgres/data_store.rs
fmarek-kindred: This is good, but we need to have Decimal instance to use "Money::from_decimal" :)

packages/cohort/src/model/json/bank_account_deserialize.rs
fmarek-kindred: It was needed for Money type,  but now I removed money and serialisation.

gk-kindred (approved):
Thanks for updating the PR

PR #57, opened 2023-07-03
feat: replicator 1 generate statemaps batch when decision is received

gk-kindred:
Improved the algorithm and the frequency at which the statemaps are send to installer.

This doesn't have metrics, but will be added in subsequent PRs.

### DIFF
--- a/packages/cohort/src/replicator/services/statemap_installer_service.rs
+++ b/packages/cohort/src/replicator/services/statemap_installer_service.rs
@@ -6,7 +6,7 @@ use log::{debug, info};
 use tokio::sync::mpsc;
 
 pub async fn installer_service<T>(
-    mut statemaps_rx: mpsc::Receiver<(Vec<StatemapItem>, Option<u64>)>,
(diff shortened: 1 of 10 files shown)


packages/cohort/src/replicator/services/statemap_installer_service.rs
```diff
@@ -16,19 +16,18 @@ where
...
-                        .unwrap();
+        if let Some(batch) = statemaps_rx.recv().await {
+            for (ver, statemap_batch) in batch {
+                debug!("[Statemap Installer Service] Received statemap batch ={statemap_batch:?} and version={ver:?}");
+                match statemap_installer.install(statemap_batch, Some(ver)).await {
+                    Ok(true) => {
+                        replicator_tx.send(ReplicatorChannel::InstallationSuccess(vec![ver])).await.unwrap();
+                    }
+                    Ok(false) => {
+                        // Do nothing if result is false.
```
fmarek-kindred, line 27: In what cases we have Ok(false) rather than Err(...) ?

fmarek-kindred (approved):
It would be good if together with this PR we get rid of replicator2 package.

PR #58, opened 2023-07-05
feat: improve commit frequency and strategy for replicator and certifier

gk-kindred:


packages/cohort/src/replicator/core.rs
```diff
@@ -182,4 +182,15 @@ where
         }
         statemaps_batch
     }
+
+    pub(crate) async fn commit_till_last_installed(&mut self) {
+        if self.last_installing > 0 {
+            if let Some(last_installed) = self.suffix.get_last_installed(Some(self.last_installing)) {
```
fmarek-kindred, line 188: minor: `Some` as param reads a bit strange here.

packages/cohort/src/replicator/services/replicator_service.rs
```diff
@@ -20,6 +20,7 @@ where
     M: MessageReciever<Message = ChannelMessage> + Send + Sync,
 {
     info!("Starting Replicator Service.... ");
+    let mut interval = tokio::time::interval(Duration::from_millis(10_000));
```
fmarek-kindred, line 23: Consider extracting into configuration property

packages/cohort/src/replicator2/cohort_replicator.rs
```diff
@@ -77,7 +77,7 @@ where
 
     pub async fn commit(&mut self) {
         if let Some(offset) = self.suffix.find_commit_offset(self.latest_in_flight) {
-            if let Err(error) = self.receiver.commit(offset).await {
+            if let Err(error) = self.receiver.commit().await {
```
fmarek-kindred, line 80: It looks like we are extracting offset from suffix into Some(offset) but then calling receiver.commit() without params. Is this intentional?

fmarek-kindred (approved):

packages/cohort/src/replicator2/cohort_replicator.rs
gk-kindred: Yes this is fine, as at the time of commit we were always using the one stored on `tpl (topic partition list)`

packages/cohort/src/replicator/services/replicator_service.rs
gk-kindred: I will update this in the new PR I am working on parallel installs, as I need to add few more environment variables in that PR. Will do it in one go.

PR #65, opened 2023-08-04
feat: Introduce opentel metrics to cohort sdk and banking example.

fmarek-kindred:


packages/metrics/src/opentel/global.rs
```diff
@@ -0,0 +1,120 @@
+use core::fmt;
```
gk-kindred, line 1: I understand why you did this, it was to pass only a reference of `meter_provider (MeterProvider)` so that you could use the `meter_provider` later to flush the last remaining metrics before exiting the app.

Although I understand the problem, I would highly recommend against keeping a custom version of the `opentelemetry::global`. Even more because we are creating a library and not an app.

This brings us the unneccessary dependency when something changes with future versions of this `global` in the official library, we have to sync it here. 

There was never needed actually, because if you look at the `MeterProvider` 

image

The properties in `MeterProvider` are `Arc`. So all you had to do is just clone the `meter_provider` in your code, before passing to the global. Because the properties in both the clones point to the same inner items. 

image

image

Here is a snippet to prove they point to the same thing
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ba9c6dd2e79a52758d3ed9826bc6fbb6

fmarek-kindred: > This brings us the unneccessary dependency when something changes with future versions of this global in the official library, we have to sync it here.

Can you please explain what dependency on global do we have? As far as I can see this is only related to original opentel through MeterProvider trait via official open tel api (look at the imports section at the top of the file).  If they change their global, then it will not affect us. 

Treat it as our version of global :) The useful part here is that we can store anything. MeterProvider is one example, the scaling factors is another. Happy to remove MeterProvider from here and apply cloning, your example suggests that it will work. But then we will end up with - for MP we use original global, for scaling - this one. Is it not better to use one for both purposes?

gk-kindred: Sorry I have to disagree again with regards to the scaling factor. I assumed the `static GLOBAL_SCALING_FACTOR` variable was available in the `opentelemetry_api::global` package as well and didn't realise that you had introduced this new variable.

I would discourage introducing a global variable just because that is more convenient. I haven't looked too much it depth, but I feel there must be better alternatives. 

- Why not set the scaling factor in a Hashmap, and use an `Arc` and pass it along to where it is required? 
     - I don't see the value of having it as a global variable rather than the convenience to access it.
  
Actually as I look deeper into the code, I would probably not recommend that as well. We know the metrics we are capturing in `cohort-sdk`. So why not let `cohort-sdk` accept the scaling factor as an input `hashmap` when it is created? Probably this would allow us to avoid using `global` altogether in the sdk? 

Why i disagree on the usage of `global`  in the current approach is because `cohort-sdk` which is a library is sharing a global variable with the binary `cohort_banking_with_sdk`, which I think isn't right

packages/metrics/Cargo.toml
```diff
@@ -5,4 +5,16 @@ edition = "2021"
 
 [dependencies]
 
-time = { version = "0.3.17" }
\ No newline at end of file
+once_cell =                 { version = "1.18.0"}
```
gk-kindred, line 8: This is not required any more, could you remove it?

packages/examples_support/src/load_generator/generator.rs
```diff
@@ -93,6 +94,8 @@ impl ControlledRateLoadGenerator {
             }
         }
 
+        tx_output.close();
```
gk-kindred, line 97: 👍

gk-kindred: Still a bit confused as to why `drop` was not working while exiting the generator. But if explicit `close` is working that should be good. 

image

packages/talos_agent/src/agent/state_manager.rs
```diff
@@ -180,6 +180,16 @@ impl<TSignalTx: Sender<Data = Signal> + 'static> StateManager<TSignalTx> {
                 decision: message.decision.clone(),
                 version: message.version,
                 safepoint: message.safepoint,
+                conflict: match message.conflicts {
```
gk-kindred, line 183: With the last PR that I merged yesterday, the certifier will only pass the conflict version under a property called `conflict_version`. Could you sync that change here?

packages/talos_agent/src/messaging/api.rs
```diff
@@ -67,9 +76,8 @@ pub struct DecisionMessage {
     #[serde(skip_deserializing)]
     pub suffix_start: u64,
     pub version: u64,
-    #[serde(skip_serializing_if = "Option::is_none")]
     pub safepoint: Option<u64>,
-    #[serde(skip_serializing_if = "Option::is_none")]
+    pub conflicts: Option<Vec<ConflictMessage>>,
```
gk-kindred, line 80: I have updated this on `certifier` end to pass `conflict_version: Option<u64>`. Could you sync that here?

packages/cohort_sdk/Cargo.toml
```diff
@@ -9,18 +9,17 @@ async-trait = { workspace = true }
...
 rdkafka-sys = { version = "4.3.0" }
 serde = { workspace = true }
 serde_json = { workspace = true }
 strum = { version = "0.25", features = ["derive"] }
+metrics = { path = "../metrics" }
 talos_agent = { path = "../talos_agent" }
-talos_suffix = { path = "../talos_suffix" }
-# talos_cohort_replicator =          { path = "../talos_cohort_replicator" }
 talos_certifier = { path = "../talos_certifier" }
 talos_certifier_adapters = { path = "../talos_certifier_adapters" }
```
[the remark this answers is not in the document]
fmarek-kindred: Oh, I missed as well. Well spotted

packages/examples_support/src/load_generator/generator.rs
fmarek-kindred: Because its arc'ed?

gk-kindred: I did check that during the review, Arc should drop the inner value if the reference count goes to 0. 

And in our use case, we don't clone it. So the reference count is 1 which on exiting will become 0, and therefore drop the inner value.

image

PR #66, opened 2023-08-09
feat: update replicator retry logic

gk-kindred:


Makefile
```diff
@@ -166,6 +166,11 @@ dev.histogram_decision_timeline_from_kafka:
 	$(call pp,histogram_decision_timeline_from_kafka...)
 	cargo run --bin histogram_decision_timeline_from_kafka --release -- $(args)
 
+## example.replicator_kafka_pg: 🧪 Runs the example replicator with installer for Kafka and Postgres
+example.replicator_kafka_pg:
```
fmarek-kindred, line 170: GK, I am a little confused. I already have this in line in master branch. Why does it come into master for the second time?

```
make
 Choose a command to run:
  withenv                                      😭 CALL TARGETS LIKE THIS `make withenv RECIPE=dev.init`
  dev.init                                     🌏 Initialize local dev environment
  dev.kafka_init                               🥁 Init kafka topic
  pg.create                                    🥁 Create database
  pg.migrate                                   🥁 Run sql on database
  pg.create_cohort                             🥁 Create database for cohort and applies DB sql
  pg.migrate_cohort                            🥁 Run sql on database for cohort
  install                                      🧹 Installs dependencies
  build                                        🧪 Compiles rust
  dev.run-agent-client                         🧪 Runs agent
  init.samply                                  🧪 Installs Samply profiler
  init.flamegraph                              🧪 Installs Flamegraph profiler
  dev.agent-run-profiler-samply                🧪 Runs agent with Samply profiler
  dev.agent-run-profiler-flamegraph            🧪 Runs agent with Flamegraph profiler
  dev.agent-run-profiler-xcode-cpu             🧪 Runs agent with XCode profiler for Time (CPU) Profiler (OSX only)
  dev.agent-run-profiler-xcode-mem             🧪 Runs agent with XCode profiler for Allocations (OSX only)
  dev.run                                      🧪 Runs rust app in watch mode
  run                                          🧪 Runs rust app
  run.release                                  🧪 Runs rust app in release mode
  run.with_mock_db                             🧪 Runs certifier with mock DB
  dev.preload_db                               🧪 Injects initial data into DB if it's not there yet
  dev.cohort_banking                           🧪 Runs Cohort with built-in replicator and executes banking transactions
  dev.cohort_banking_with_sdk                  🧪 Runs an example of rust app "Cohort Banking" which use cohort_sdk.
  dev.histogram_decision_timeline_from_kafka   🧪 Reads all decisions from kafka and prints processing timeline as csv
  example.replicator_kafka_pg                  🧪 Runs the example replicator with installer for Kafka and Postgres
  dev.run_replicator                           🧪 Runs replicator
  lint                                         🧹 Checks for lint failures on rust
  test.unit                                    🧪 Runs unit tests
  test.some-unit                               🧪 Runs specified unit tests
  test.unit.coverage                           🧪 Runs rust unit tests with coverage 'cobertura' and 'junit' reports
```

examples/cohort_banking_with_sdk/examples/cohort_banking_with_sdk.rs
```diff
@@ -205,39 +198,6 @@ async fn main() -> Result<(), String> {
     Ok(())
 }
 
-fn start_queue_monitor(queue: Arc<Receiver<TransferRequest>>) -> JoinHandle<Result<(), String>> {
```
fmarek-kindred, line 208: Why we are removing this?

Makefile
gk-kindred: Aaah, that is because it was not rebased on `master`. I just pushed after rebase so it shouldn't be there anymore

examples/cohort_banking_with_sdk/examples/cohort_banking_with_sdk.rs
gk-kindred: It is not really needed. This was useful when replicator was part of `cohort-sdk`.  I forgot to remove it in the earlier PR which was merged to master last week.

packages/cohort_banking/src/callbacks/statemap_installer.rs
```diff
@@ -1,68 +1,126 @@
...
+    async fn install(&self, statemap: Vec<StatemapItem>, snapshot_version: u64) -> Result<ReplicatorInstallStatus, String> {
+        let mut bank_transfer_error_retry_check = retry_allowed_on_error(self.max_retry);
+        let mut snapshot_retry_check = retry_allowed_on_error(self.max_retry);
 
-                let updated_rows = tx.execute(sql, params).await.map_err(|e| e.to_string())?;
+        let mut cnn = self.database.get().await.map_err(|e| e.to_string())?;
+        loop {
+            let tx = cnn.transaction().await.map_err(|e| e.to_string())?;
+            if !statemap.is_empty() {
+                let updated_rows_res = BankStatemapInstaller::update_bank_transfer_request(&tx, &statemap, snapshot_version).await;
```
fmarek-kindred, line 85: I suggest to rename this method "update_bank_transfer_request". Be reading it feels like we are updating the request. However we are installing the given state map.

packages/talos_cohort_replicator/src/services/statemap_installer_service.rs
```diff
@@ -11,13 +14,72 @@ pub struct StatemapInstallerConfig {
     pub thread_pool: Option<u16>,
 }
 
+async fn statemap_install_future(
```
fmarek-kindred, line 17: RE: method name with ending _future. What does it mean? When reading it is not straightforward to understand what will happen when method is called.

packages/cohort_banking/src/callbacks/statemap_installer.rs
gk-kindred: will update this.

examples/cohort_banking_with_sdk/examples/cohort_banking_with_sdk.rs
fmarek-kindred: This one is observing how quickly cohort is draining the generated queue. It is not related to replicator. I prefer it to stay :) What you saw previously is that monitor was printing stats of generator queue and replicator queue (via heartbeat channel), but now it observes only generator. Btw if we remove it, I am not sure how system will stop itself, because async services are running forever, the don't know when to stop.

gk-kindred: aaah okay, I must have got confused with the earlier logic. Will add this back. 

The app stops correctly even after I remove it. In fact one reason why I removed it was, sometimes, the monitor would preemptively stop the app even before it reached the target number of `txn`. Probably there is a bug somewhere there, which I didn't dig too much in detail, but just deleted the code :)

packages/cohort_banking/src/callbacks/statemap_installer.rs
```diff
@@ -1,68 +1,106 @@
...
+    END),
+    "version" = ($4)::BIGINT
+    WHERE ba."number" IN (($1)::TEXT, ($2)::TEXT)
+    AND ba."version" < ($4)::BIGINT
+    "#;
+
+const SNAPSHOT_UPDATE_QUERY: &str = r#"UPDATE cohort_snapshot SET "version" = ($1)::BIGINT WHERE id = $2 AND "version" < ($1)::BIGINT"#;
+
+fn is_retriable_error(error: &str) -> bool {
+    // TODO: improve retriable error detection when error moves from String to enum or struct.
```
fmarek-kindred, line 25: Agree, this is Postgres text. Somehow it should be abstracted depending on DB type.

packages/cohort_banking/src/callbacks/statemap_installer.rs
```diff
@@ -1,68 +1,106 @@
...
-        loop {
-            // from      = 1
-            // to        = 2
-            // amount    = 3
-            // new_ver   = 4
-
-            let mut cnn = self.database.get().await.map_err(|e| e.to_string())?;
-            let tx = cnn.transaction().await.map_err(|e| e.to_string())?;
+impl BankStatemapInstaller {
+    async fn install_bank_transfer_statemap(tx: &Transaction<'_>, statemap: &[StatemapItem], _snapshot_version: u64) -> Result<u64, String> {
```
fmarek-kindred, line 36: thanks

packages/cohort_banking/src/callbacks/statemap_installer.rs
```diff
@@ -1,68 +1,106 @@
...
+        loop {
+            let tx = cnn.transaction().await.map_err(|e| e.to_string())?;
+            if !statemap.is_empty() {
+                let updated_rows_res = BankStatemapInstaller::install_bank_transfer_statemap(&tx, &statemap, snapshot_version).await;
+
+                match updated_rows_res {
+                    Ok(updated_rows) => {
+                        if updated_rows > 0 {
+                            log::debug!(
+                                "No rows were updated when installing: {:?}. Snapshot will be set to: {}",
```
fmarek-kindred, line 78: Is this typo? Says "No rows"

packages/talos_cohort_replicator/src/services/statemap_installer_service.rs
```diff
@@ -11,13 +14,55 @@ pub struct StatemapInstallerConfig {
...
+        }
+
+        Err(err) => {
+            error!(
+                "Installed failed for version={version:?} with time={:?} error={err:?}",
+                start_installation_time.elapsed()
+            );
+            replicator_tx
+                .send(ReplicatorChannel::InstallationFailure(format!(
+                    "Installed failed for version={version:?} error={err:?}"
```
fmarek-kindred, line 43: Is it a typo in "Installed"?

fmarek-kindred (approved):
Looks good. Left minor typo warnings.

packages/talos_cohort_replicator/src/services/statemap_installer_service.rs
gk-kindred: Isn't `future` is pretty self explanatory in this context. 

To me it was pretty obvious based on being used in the tokio task, that this `fn` is the future expected in the tokio task.

packages/cohort_banking/src/callbacks/statemap_installer.rs
gk-kindred: Aah I never paid attention to this. Good catch. I just moved around the original if block that you had written. I will update this.

image

packages/talos_cohort_replicator/src/services/statemap_installer_service.rs
gk-kindred: yup will update this

packages/talos_cohort_replicator/src/services/statemap_installer_service.rs
fmarek-kindred: To me method name is about "do something" or "get something", "parse something", "create something" and etc From that context I don't know what _future  does here. Why do we need to encode a data type into the method name :)

PR #71, opened 2023-08-17
feat: Add abort reason to CapturedState.

fmarek-kindred:
This is for https://jira.kindredgroup.com/browse/KDT-72


packages/cohort_sdk/src/cohort.rs
```diff
@@ -380,6 +388,10 @@ impl Cohort {
             Ok(local_state) => local_state,
         };
 
+        if let Some(reason) = local_state.abort_reason {
```
gk-kindred, line 392: Could you check the value passed to `abort_reason`? 

I had a look at the functions `get_state_using_one_query` and `get_state_using_two_queries`, and both set the `abort_reason` to `None`. So this `if` statement will never be true.

gk-kindred (approved):

anil0906 (approved):

PR #75, opened 2023-08-23
feat: get snapshot and certification request payload via callback

gk-kindred:


packages/cohort_banking/src/app.rs
```diff
@@ -82,23 +80,24 @@ impl Handler<TransferRequest> for BankingApp {
...
 
         let oo_inst = OutOfOrderInstallerImpl {
             database: Arc::clone(&self.database),
-            request: request_copy,
             detailed_logging: false,
             counter_oo_no_data_found: Arc::clone(&self.counter_oo_no_data_found),
             single_query_strategy,
         };
 
+        let oo_installer_callback = |payload: OOOInstallerPayload| oo_inst.install(payload);
```
fmarek-kindred, line 94: can it be a reference to function? `oo_inst::install`

packages/cohort_banking/src/callbacks/oo_installer.rs
```diff
@@ -63,10 +61,17 @@ impl OutOfOrderInstallerImpl {
         }
     }
 
-    async fn install_using_polling(&self, _xid: String, safepoint: u64, new_version: u64, _attempt_nr: u32) -> Result<OutOfOrderInstallOutcome, String> {
+    async fn install_using_polling(
```
fmarek-kindred, line 64: can be deleted

packages/cohort_banking/src/callbacks/oo_installer.rs
```diff
@@ -163,7 +165,7 @@ impl OutOfOrderInstallerImpl {
 
         // Quickly grab the snapshot to check whether safepoint condition is satisfied. Any row can be used for that.
         let (_, _, _, snapshot) = &result[0];
-        if (*snapshot as u64) < safepoint {
+        if &(*snapshot as u64) < safepoint {
```
fmarek-kindred, line 168: are we comparing reference? Why `&` ?

packages/cohort_banking/src/callbacks/state_provider.rs
```diff
@@ -118,18 +131,35 @@ impl StateProviderImpl {
...
+
+        let mut accounts: Vec<&(dyn ToSql + Sync)> = vec![];
+        request.candidate.readset.iter().for_each(|x| accounts.push(x));
+
+        let state = if self.single_query_strategy {
+            self.get_state_using_one_query(&accounts).await
         } else {
-            self.get_state_using_two_queries().await
-        }
+            self.get_state_using_two_queries(&accounts).await
```
fmarek-kindred, line 149: I don't like this part where signature is dynamic accounts array. In this app we will have 2 accounts as fixed structure. It is also used like that in SQL query. Masking it into param which accepts a dynamic array looks confusing.

packages/cohort_sdk/src/cohort.rs
```diff
@@ -144,45 +143,108 @@ impl Cohort {
...
+        let response = self.send_to_talos(create_certifier_candidate_callback).await?;
         let span_1_val = span_1.elapsed().as_nanos() as f64 / 1_000_000_f64;
 
         let h_talos = Arc::clone(&self.talos_histogram);
         tokio::spawn(async move {
             h_talos.record(span_1_val * 100.0, &[]);
         });
 
-        if response.decision == Decision::Aborted {
+        if response.decision == Decision::Aborted || response.statemaps.is_none() {
```
fmarek-kindred, line 211: yes, good catch

packages/cohort_sdk/src/cohort.rs
```diff
@@ -144,45 +143,108 @@ impl Cohort {
...
             return Ok(response);
         }
 
-        // system error if we have Commit decision but no safepoint is given
-        let safepoint = response.safepoint.unwrap();
-        let new_version = response.version;
+        let oooinstall_payload = OOOInstallerPayload {
+            xid: response.xid.clone(),
+            version: response.version,
+            safepoint: response.safepoint.unwrap(),
```
fmarek-kindred, line 218: is unwrap safe here?

packages/cohort_sdk/src/cohort.rs
```diff
@@ -144,45 +143,108 @@ impl Cohort {
...
+            statemaps: response.statemaps.clone().unwrap(),
```
fmarek-kindred, line 219: is unwrap safe here?

packages/cohort_sdk/src/cohort.rs
```diff
@@ -315,32 +380,12 @@ impl Cohort {
...
-                    false
                 }
-            };
-
-            let rslt_response = result.unwrap();
-            if is_success || self.config.retry_attempts_max <= attempts {
-                break rslt_response;
             }
 
-            match rslt_response {
```
fmarek-kindred, line 327: debug messages disappeared :)

packages/cohort_sdk/src/cohort.rs
```diff
@@ -360,49 +405,40 @@ impl Cohort {
...
             Err(SnapshotPollErrorType::FetchError { reason }) => return CertificationAttemptOutcome::DataError { reason },
             Err(SnapshotPollErrorType::Timeout { waited }) => {
                 return CertificationAttemptOutcome::SnapshotTimeout {
                     waited,
                     conflict: previous_conflict.unwrap(),
                 }
             }
-            Ok(local_state) => local_state,
+            Ok(CertificationCandidateCallbackResponse::Cancelled(reason)) => {
+                return CertificationAttemptOutcome::ClientAborted { reason };
```
fmarek-kindred, line 427: There are too many changes in this class, I might be getting confused here. Aren't we doing something like this? From Cancelled to ClientAborted and then again to Cancelled?  If yes, maybe we should use single naming approach, either all Cancelled or all ClientAborted :)

packages/cohort_sdk/src/cohort.rs
```diff
@@ -436,79 +473,39 @@ impl Cohort {
...
-                                if waited >= timeout {
-                                    return Err(SnapshotPollErrorType::Timeout { waited });
-                                }
-                                delay_controller.sleep().await;
-                                continue;
-                            } else {
-                                break Ok(current_state);
-                            }
-                        }
+        let conflict = previous_conflict.unwrap_or(0);
```
fmarek-kindred, line 486: I was wondering how did you flatten the Optionals, here is the answer, a bit of cheating  with 0 :)

fmarek-kindred:
@gk-kindred  I left a few comment here and there. In this example I see how the same functionality can be coded using closures instead of Traits. But honestly not sure what actually improved by doing that. Method signatures became harder to read:

Before:
```
pub async fn certify<S, O>(...)
where
        S: ItemStateProvider,
        O: OutOfOrderInstaller,
        
let install_result = oo_installer.install(...) // easy to understand what will happen.
let result_local_state = state_provider.get_state().await;
```

After:

```
pub async fn certify<F, Fut, IF, IFut>(...)
where
        F: Fn() -> Fut,
        Fut: Future<Output = Result<CertificationCandidateCallbackResponse, String>>,
        IF: Fn(OOOInstallerPayload) -> IFut,
        
let install_result = installer_callback(install_payload.clone()).await; // just reading "installer_callback" is not clear what is going on.
// this comment is not related to trait vs closure, just noticed we forget to rename. I think we are not loading state anymore, but rather making a new request every time? 
let result_local_state = get_state_callback_fn().await;
```

packages/cohort_banking/src/callbacks/oo_installer.rs
gk-kindred: aaah that was left over of something I was trying and forgot to clean up. Thanks for noticing this, I will update this.

packages/cohort_sdk/src/cohort.rs
gk-kindred: Yes ideally it should be safe, as we reach here only if we have committed messages with statemaps to install. 

But to make it safer, just in case there is a bug in certifier, I am update the code above as:-

```rust
if response.safepoint.is_none() || response.statemaps.is_none() {
            return Ok(response);
        }
```

packages/cohort_sdk/src/cohort.rs
gk-kindred: Answered in the previous commit

packages/cohort_banking/src/callbacks/state_provider.rs
gk-kindred: agree, I was experimenting something and just left it there, I have modified this now.

packages/cohort_sdk/src/cohort.rs
gk-kindred: that is a safe cheating :)

packages/cohort_sdk/src/cohort.rs
gk-kindred: yes, that got cleaned up by mistake. I have added it back with some minor changes

packages/cohort_sdk/src/cohort.rs
```diff
@@ -144,45 +143,108 @@ impl Cohort {
...
-        oo_installer: &O,
+        create_certifier_candidate_callback: &F,
+        installer_callback: &IF,
     ) -> Result<model::CertificationResponse, ClientError>
     where
-        S: ItemStateProvider,
-        O: OutOfOrderInstaller,
+        F: Fn() -> Fut,
+        Fut: Future<Output = Result<CertificationCandidateCallbackResponse, String>>,
+        IF: Fn(OOOInstallerPayload) -> IFut,
```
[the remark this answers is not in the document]
gk-kindred: We do need the `payload`. Its this struct:-

```rust
#[derive(Debug, Clone)]
pub struct OOOInstallerPayload {
    pub xid: String,
    pub version: u64,
    pub safepoint: u64,
    pub statemaps: Vec<HashMap<String, Value>>,
}
```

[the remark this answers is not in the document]
gk-kindred: I renamed it to `OutOfOrderInstallRequest`

packages/cohort_banking/src/app.rs
gk-kindred: can I ask why you want to do that in the above example?

packages/cohort_sdk/src/cohort.rs
gk-kindred: good catch, that should be cancelled. I missed that `ClientAbort` when I changed the rest. Will update it

packages/cohort_banking/src/app.rs
fmarek-kindred: It's a question, GK. Not suggesting to change anything. But if we can use reference maybe it is shorter that way?

packages/cohort_sdk/src/cohort.rs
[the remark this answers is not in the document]
fmarek-kindred: Now I know what it does without looking at the code where it is used. Good name!

gk-kindred:
> @gk-kindred I left a few comment here and there. In this example I see how the same functionality can be coded using closures instead of Traits. But honestly not sure what actually improved by doing that. Method signatures became harder ...

@fmarek-kindred - Regarding this I partially agree. Probably I went overboard by changing it for `installer` as well, where it was not required. But for getting the `certification request`, I feel it is better to have it as I have created as it makes it more flexible.

See this example, where it is up to the user how they create the function, closure or method in a class/struct to get the `request`:-
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e3b02b3bb3e69b69d1ee233449acc38c

That level of flexibility is had to get using just `generic with traits`.

I have made the changes in my local for putting the `installer` as `generic using trait`, but I would like to keep the `request` part as it's implemented currently. Let me know if this is fine

packages/cohort_sdk/src/cohort.rs
ekoutanov: If there's a bug in the certifier, how can we even safely act on a decision it provided?

packages/cohort_sdk/src/model/mod.rs
```diff
@@ -42,19 +29,21 @@ pub struct ResponseMetadata {
...
 }
 
-#[derive(strum::Display)]
+#[derive(strum::Display, Debug)]
 // this is napi friendly copy of talos_agent::agent::errors::AgentErrorKind
 pub enum ClientErrorKind {
     Certification,
     CertificationTimeout,
-    ClientAborted,
+    SnapshotTimeout,
```
fmarek-kindred, line 37: Is this value used anywhere? Three timeouts here look suspicious :)

gk-kindred: updated. Thanks for spotting this.

fmarek-kindred (approved):

PR #80, opened 2023-09-18
feat: Errors handling between Rust and JS

fmarek-kindred:
The main ideas:

Throwing error

- Rust code should always "throw" some Talos error, in our case all public method are throwing `ClientError` so, there was no need to do any changes here.
- Create simple "napi-managed" enum `SdkErrorKind` in RS and map its values to `ClientError.kind`, in our case that is `ClientErrorKind.rs` enum.
- In the napi adapter layer "cohort_sdk_js" declare a simple container class `SdkErrorContainer" whose purpose is to collect  attributes from `ClientError` and make them a) serialisable to JSON b) transferrable to JS layer
- When `cohort_sdk_js` is handling rust error it packs data into `SdkErrorContainer`, serialises it to json + string and sets into `napi::Error.message`

Receiving and mapping error

- JS caller will receive this error as GenericError with message looking like JSON object.
- Introduce `cohort_sdk_client` JS project which wraps every public service class from `cohort_sdk_js`. For example, `cohort_sdk_js` exposes logical service "initiator". We program this service as `cohort_sdk_js::InternalInitiator` and do not let JS client to use it directly. Instead, we wrap this class into "cohort_sdk_client::Initiator" so that every public method is a delegate with `try .. catch.` When catching error we catch a normal JS error, use it's `message` attribute to parse into `SdkErrorContainer`. Then we re-throw resulting info as "official" `TalosSdkError`. This is JS class extending JS Error with enum inside it.

The end client only aware of: 

- `cohort_sdk_client` as dependency
- `cohort_sdk_client::Initiator` as service class, 
- `cohort_sdk_client::TalosSdkError` as error which will be thown from service classes. It has `TalosSdkError.kind` as `SdkErrorKind` enum.


cohort_banking_initiator_js/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@types/pg": "^8.10.2",
+    "cohort_sdk_client": "file:../cohort_sdk_client",
     "cohort_sdk_js": "file:../packages/cohort_sdk_js",
```
fmarek-kindred, line 21: If we consider only classes then this dependency is not required, but I had to add it here because package delivers not only public classes but also darwin image.

cohort_banking_initiator_js/src/banking-app.ts
```diff
@@ -25,13 +25,14 @@ export class BankingApp {
 
     async init() {
         this.initiator = await Initiator.init(sdkConfig)
```
fmarek-kindred, line 27: If kafka is down the error will bubble into main method. There is an example how we can handle typed errors.

cohort_banking_initiator_js/src/main.ts
```diff
@@ -71,6 +111,16 @@ new Promise(async (resolve) => {
         queue,
         fnFinish,
     )
-    await app.init()
+    try {
```
fmarek-kindred, line 114: this is example where caller can deal with specific errors. I skipped else statement here.

cohort_sdk_client/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+    SdkErrorKind,
+ } from "cohort_sdk_js"
+
+class TalosSdkError extends Error {
+    constructor(readonly kind: SdkErrorKind, readonly message: string, options?: ErrorOptions) {
+        super(message, options)
+    }
+}
+
+export {
```
fmarek-kindred, line 15: Export only what we allow client to use.

cohort_sdk_client/src/initiator.ts
```diff
@@ -0,0 +1,35 @@
+import { InternalInitiator, JsInitiatorConfig, OutOfOrderRequest, SdkErrorKind } from "cohort_sdk_js"
+import { isSdkError } from "./internal"
+import { TalosSdkError } from "."
+
+export class Initiator {
```
fmarek-kindred, line 5: Wrapper with error handler.

cohort_sdk_client/src/internal.ts
```diff
@@ -0,0 +1,5 @@
+import { SDK_CONTAINER_TYPE } from "cohort_sdk_js"
```
fmarek-kindred, line 1: The value is managed in the Rust layer.

cohort_sdk_client/src/internal.ts
```diff
@@ -0,0 +1,5 @@
...
+
+export const isSdkError = (reason: string): boolean => {
+    return (reason && reason.indexOf(`"typ":"${ SDK_CONTAINER_TYPE }"`) > 0 && reason.startsWith("{") && reason.endsWith("}"))
```
fmarek-kindred, line 4: Tried to use "_typ" as name but napi seems renaming it to "typ" anyway.

packages/cohort_sdk_js/src/initiator/mod.rs
```diff
@@ -108,33 +109,29 @@ pub struct JsCertificationCandidateCallbackResponse {
...
+    ) -> napi::Result<()> {
         let new_request_provider = NewRequestProvider { make_new_request_callback };
         let ooo_impl = OutOfOrderInstallerImpl { ooo_callback };
-        // println!("Initiator.certify(): invoking cohort.certify(...)");
         let make_new_request = || new_request_provider.make_new_request();
-        let _res = self.cohort.certify(&make_new_request, &ooo_impl).await.map_err(map_error_to_napi_error)?;
-
-        // println!("Initiator.certify(): after cohort.certify(...)");
-        Ok("Success".to_string())
+        let _res = self.cohort.certify(&make_new_request, &ooo_impl).await.map_err(map_error)?;
```
fmarek-kindred, line 133: That's where we use out error handling and packaging logic.

packages/cohort_sdk_js/src/typed_errors_test.rs
```diff
@@ -0,0 +1,25 @@
+use napi_derive::napi;
+
+use crate::sdk_errors::{SdkErrorContainer, SdkErrorKind};
+
+#[napi]
+pub struct InternalSomeRustServiceClass {}
+
+#[napi]
+impl InternalSomeRustServiceClass {
```
fmarek-kindred, line 9: Ignore this guy.

cohort_banking_initiator_js/package.json
anil0906: You can add an optional dependency section in cohort_sdk_js. So based on the platform it will get the required dependency. From a consumer perspective, they only need to add cohort_sdk_client as a dependency.

https://github.com/kindredgroup/scylla/blob/feature/add-missing-dependencies/scylla_pg_js/package.json#L39-L43

gk-kindred (approved):
Nice 👍

PR #82, opened 2023-09-22
feat: Optimise replicator commit logic

fmarek-kindred:
- Fix for offset lag in the replicator kafka group
- Fix for too frequent commits in certifier
- Commit change from sync to async.


gk-kindred (approved):
Thanks for fixing this :-)

PR #83, opened 2023-09-25
feat: Talos messenger base

gk-kindred:
Future PRs would implement:-

- Pruning of suffix
- Committing offsets to Kafka
- More unit tests - especially around suffix used in messenger


examples/certifier_kafka_pg/examples/certifier_kafka_pg.rs
```diff
@@ -18,7 +18,21 @@ async fn main() -> Result<(), impl std::error::Error> {
...
     info!("Talos certifier starting...");
 
-    let kafka_config = KafkaConfig::from_env(None);
+    let mut kafka_config = KafkaConfig::from_env(None);
+    kafka_config.extend(
+        None,
+        Some(
+            [
+                ("enable.auto.commit".to_string(), "false".to_string()),
+                ("auto.offset.reset".to_string(), "earliest".to_string()),
```
fmarek-kindred, line 27: Is this push intentional?

examples/messenger_using_kafka/src/kafka_producer.rs
```diff
@@ -0,0 +1,48 @@
...
+    C: ProducerContext<DeliveryOpaque = Box<MessengerProducerDeliveryOpaque>> + 'static,
+{
+    type Payload = KafkaAction;
+    type AdditionalData = u32;
+    fn get_publish_type(&self) -> PublishActionType {
+        PublishActionType::Kafka
+    }
+
+    async fn send(&self, version: u64, payload: Self::Payload, additional_data: Self::AdditionalData) -> () {
+        info!("[MessengerKafkaPublisher] Publishing message with payload=\n{payload:#?}");
```
fmarek-kindred, line 31: It's good if we could agree on prefix, so that we use it consistently. Somewhere in the code I used format like this: "MessengerKafkaPublisher.send(): ...". Does not matter really just better when there is one format :)

examples/messenger_using_kafka/src/kafka_producer.rs
```diff
@@ -0,0 +1,48 @@
...
+        let payld = std::str::from_utf8(&bytes).unwrap();
+        info!("[MessengerKafkaPublisher] base_record=\n{payld:#?}");
+
+        let delivery_opaque = MessengerProducerDeliveryOpaque {
+            version,
+            total_publish_count: additional_data,
+        };
+
+        self.publisher
+            .publish_to_topic("test.messenger.topic", "test", payld, None, Box::new(delivery_opaque))
```
fmarek-kindred, line 45: I assume this hardcoded topic name will be moved somewhere in the future work, is that right?

packages/talos_messenger_actions/src/kafka/producer.rs
```diff
@@ -0,0 +1,143 @@
...
+                info!("Message {:?} {:?}", msg.key(), msg.offset());
+                // TODO: GK - what to do on error? Panic?
+                let _ = block_on(self.tx_feedback_channel.send(MessengerChannelFeedback::Success(
+                    version,
+                    "kafka".to_string(),
+                    delivery_opaque.total_publish_count,
+                )));
+            }
+            Err(err) => {
+                // TODO: GK - what to do on error? Panic?
```
fmarek-kindred, line 51: If we panic and do not commit offset then pods will keep restarting and will be stuck in that consume/panic/restart loop.

packages/talos_messenger_actions/src/kafka/producer.rs
```diff
@@ -0,0 +1,143 @@
...
+impl MessagePublisher for KafkaProducer {
+    async fn publish_message(&self, key: &str, value: &str, headers: Option<HashMap<String, String>>) -> Result<(), SystemServiceError> {
+        let record = BaseRecord::to(&self.topic).payload(value).key(key);
+
+        let record = match headers {
+            Some(x) => record.headers(build_kafka_headers(x)),
+            None => record,
+        };
+
+        debug!("Preparing to send the Decision Message. ");
```
fmarek-kindred, line 119: Decision?

packages/talos_messenger_core/src/core.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::errors::MessengerServiceResult;
+
+#[derive(Debug, Display, Serialize, Deserialize, EnumString, EnumIter, Clone, Eq, PartialEq)]
+pub enum CommitActionType {
+    #[strum(serialize = "publish")]
+    Publish,
+}
+
+#[derive(Debug, Display, Serialize, Deserialize, EnumString, Clone, Eq, PartialEq)]
+pub enum PublishActionType {
```
fmarek-kindred, line 16: How about we rename this to represent a list of brokers? If I read the intention correctly this enum wants to list what messaging brokers we can publish to. In that case its no so much of a "...ActionType" anymore.

packages/talos_messenger_core/src/core.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#[derive(Debug)]
+pub struct MessengerCommitActions {
+    pub version: u64,
+    pub commit_actions: HashMap<String, Value>,
+}
+
+pub enum MessengerChannelFeedback {
+    Error(u64, String),
```
fmarek-kindred, line 44: Why not to make simple structs or named variables. Tuple style allows only to show types but names and purposes are very not clear :)

packages/talos_messenger_core/src/suffix.rs
```diff
@@ -0,0 +1,297 @@
...
+        } else {
+            None
+        }
+    }
+
+    fn get_last_installed(&self, _to_version: Option<u64>) -> Option<&SuffixItem<T>> {
+        todo!()
+    }
+
+    // TODO: GK - Elevate this to core suffix
```
fmarek-kindred, line 207: good idea

packages/talos_messenger_core/src/talos_messenger_service.rs
```diff
@@ -0,0 +1,46 @@
...
+
+pub struct TalosMessengerService {
+    pub services: Vec<Box<dyn MessengerSystemService + Send + Sync>>,
+}
+
+impl TalosMessengerService {
+    pub async fn run(self) -> MessengerServiceResult {
+        let service_handle = self.services.into_iter().map(|mut service| tokio::spawn(async move { service.run().await }));
+
+        let k = try_join_all(service_handle).await.map_err(|e| MessengerServiceError {
```
fmarek-kindred, line 16: Do we need this join all? By reading I understand there is only one `service_handle`.

fmarek-kindred (approved):
Left some minor comments. Can merge to reduce the number of pending PRs, but will need to finish bits a pieces in TODOs

packages/talos_messenger_actions/src/kafka/producer.rs
ekoutanov: What you're alluding to is that the code should be generic and not only deal with decisions, right?

packages/talos_messenger_actions/src/kafka/producer.rs
ekoutanov: Failure to commit, while undesirable, shouldn't violate safety.

packages/talos_messenger_core/src/talos_messenger_service.rs
ekoutanov: The tasks are mapped from an iterator, so probably the variable should be `service_handles` (plural).

examples/certifier_kafka_pg/examples/certifier_kafka_pg.rs
gk-kindred: yes that is intentional, as I need that till I enable the manual commit. This helps me in consuming the same set of messages from Kafka and run through my new changes in code.

examples/messenger_using_kafka/src/kafka_producer.rs
gk-kindred: oh well spotted, I completely missed this. It shouldn't be hardcoded anymore. I will fix that in this PR

packages/talos_messenger_actions/src/kafka/producer.rs
gk-kindred: Yes, that should be a generic message.

I will fix that, thanks for pointing that out. 

This impl of `MessagePublisher` is actually copy from the producer I created for certifier. 

The difference between this `producer` vs the one from certifier is that, this gets more generic, allowing to have custom `delivery context` and `delivery opaque`. 

Once this going through some clean up, I would like to move this into the `certifier` producer and use it here.

packages/talos_messenger_core/src/core.rs
gk-kindred: Actually can I leave this comment for the next PR please? Some part of it actually would get more generic. Like `line 10 - CommitActionType` is not used anymore.  And my idea around the implementation has become bit more generic since the time these enums where created.

packages/talos_messenger_core/src/core.rs
gk-kindred: Agree, just the lazy me to blame :) I will update this

packages/talos_messenger_core/src/talos_messenger_service.rs
gk-kindred: yes, should be `service_handles`

examples/certifier_kafka_pg/examples/certifier_kafka_pg.rs
gk-kindred: oh sorry, just realised that in the `certifier`, I think this is part of the commit before your PR to fix the auto commit went it. I will remove this now.

packages/talos_messenger_actions/src/kafka/producer.rs
gk-kindred: The `todo` is for the channel. What to do if the channel gives an error. As this is the way I get to know some Kafka action has been processed, and then mark it in suffix. 

Realistically the channel would give error when we try to send the message after the receiver was closed, so can I make the assumption that we can ignore the error check for `channel.send` ?

packages/talos_messenger_core/src/core.rs
gk-kindred: Sorry, I missed this one while updating the review comments changes. Will update it in the next PR

PR #84, opened 2023-09-29
feat: talos messenger suffix prune and message receiver offset commit

gk-kindred:


packages/talos_certifier/src/ports/message.rs
```diff
@@ -16,7 +16,8 @@ pub trait MessageReciever: SharedPortTraits {
     async fn subscribe(&self) -> Result<(), SystemServiceError>;
     async fn commit(&self) -> Result<(), SystemServiceError>;
     fn commit_async(&self) -> Option<JoinHandle<Result<(), SystemServiceError>>>;
-    async fn update_savepoint(&mut self, offset: i64) -> Result<(), SystemServiceError>;
+    fn update_savepoint(&mut self, offset: i64) -> Result<(), Box<SystemServiceError>>;
+    async fn update_savepoint_async(&mut self, offset: i64) -> Result<(), SystemServiceError>;
```
fmarek-kindred, line 20: did you mean safepoint? "f" letter vs "v" ?

fmarek-kindred (approved):

packages/talos_certifier/src/ports/message.rs
gk-kindred: When I wrote the function initially, I was thinking of something like a DB Savepoint. But that is not correct. I will rename that in the next PR. Thanks for pointing this out.

PR #85, opened 2023-10-09
feature: Publish cohort SDK NPSs with bundled rust src

fmarek-kindred:


.github/workflows/npm.yml
```diff
@@ -0,0 +1,120 @@
+name: cohort_sdk_js
```
fmarek-kindred, line 1: This workflow does not work and it is not activated

.github/workflows/npm_publish_light.yml
```diff
@@ -0,0 +1,126 @@
+name: Build and Publish Cohort SDK NPMs
+on:
+    #workflow_call:
+    push:
```
fmarek-kindred, line 4: Now triggered on push, but after PR will change to be on manual button click.

packages/cohort_sdk_js/npm/darwin-arm64/package.json
```diff
@@ -0,0 +1,18 @@
+{
```
fmarek-kindred, line 1: files in npm/ are not used.

scripts/github-actions-ci/set-npm-dev-version.sh
```diff
@@ -0,0 +1,39 @@
...
+#   $1 - Required. The package-name of node module
+#   $2 - Optional. The parameter name under which the resulting value will be exported into
+#        $GITHUB_OUTPUT
+# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+selectNodeName=$1
+outputParamName=$2
+
+if [ "$selectNodeName" == "" ]; then
+    echo "Missing the furst parameter - npm package name"
```
gk-kindred, line 19: small typo `furst` -> `first`

packages/cohort_sdk_js/package.json
```diff
@@ -25,14 +28,23 @@
...
+    "index.js",
+    "cohort_sdk_js*",
+    "scripts/postinstall.sh"
+  ],
   "scripts": {
-    "artifacts": "napi artifacts",
+    "artifacts": "napi artifacts -d .",
     "build": "napi build --platform --release",
     "build:debug": "napi build --platform",
     "test": "ava",
```
gk-kindred, line 42: Hmm, interesting, haven't seen `ava` in years. 

Just wanted to know why you why choose `ava` over some of the other popular testing libraries like `jest`, `mocha`

.github/workflows/npm_publish_light.yml
```diff
@@ -0,0 +1,126 @@
...
+
+        name: Publish SDK NPS with bundled Rust for node@16
+        runs-on: ${{ matrix.settings.host }}
+        steps:
+            - uses: actions/checkout@v3
+
+            - name: Setup node
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 16
```
gk-kindred, line 29: Probably use `16` and `18`

.github/workflows/npm_publish_light.yml
```diff
@@ -0,0 +1,126 @@
...
+        strategy:
+            fail-fast: false
+            matrix:
+                settings:
+                  - host: macos-latest
+                    dirSdkJs: packages/cohort_sdk_js
+                    dirSdkJsClient: cohort_sdk_client
+                    dirCohortInitiator: cohort_banking_initiator_js
+                    dirCohortReplicator: cohort_banking_replicator_js
+                    customiseHostCommand: echo "No customizations needed"
```
gk-kindred, line 19: I probably wouldn't pass custom commands over a variable and inject it. 

Two reasons:- 
- This feels a bit unsafe to me. 
- Also when I look at the actions it makes it harder to interpret the workflow when custom commands come from variables. (I referred to `npm.yml` file and that is why this comment)

cohort_banking_initiator_js/README.md
```diff
@@ -0,0 +1,121 @@
+# Cohort Initiator JS App
```
gk-kindred, line 1: Nice 👍 👍  Thanks for writing this

packages/cohort_sdk_js/package.json
fmarek-kindred: I did not, it was copy from scylla. In fact no testing is done in this repo. Let me remove it.

.github/workflows/npm_publish_light.yml
fmarek-kindred: We have ticket to do testing with node 18. If we are going to support both then the whole workflow should be different in the way that it should be publishing two artefacts, one for v16 and one for v18. I prefer to have that work done as separate ticket (if we need to do it)

gk-kindred (approved):
Thanks for updating 👍

PR #86, opened 2023-10-09
feat: final tweaks to messenger

gk-kindred:
- Some refactors and renames
- Change how whitelist of actions are fed to the messenger service
- Move messenger suffix related `get_mut` and `get_meta` to the core suffix.
- Handle how actions failed to process will be marked in the suffix

### DIFF
--- a/packages/talos_messenger_actions/src/kafka/models.rs
+++ b/packages/talos_messenger_actions/src/kafka/models.rs
@@ -2,14 +2,41 @@ use ahash::HashMap;
 use serde::{Deserialize, Serialize}; // 1.0.130
 use serde_json::{self};
 
+fn default_text_plain_encoding() -> String {
+    "text/plain".to_string()
+}
+
+fn default_application_json_encoding() -> String {
+    "application/json".to_string()
+}
+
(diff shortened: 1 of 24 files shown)


packages/talos_messenger_actions/src/kafka/models.rs
```diff
@@ -2,12 +2,41 @@ use ahash::HashMap;
 use serde::{Deserialize, Serialize}; // 1.0.130
 use serde_json::{self};
 
+fn default_text_plain_encoding() -> String {
+    "text/plain".to_string()
```
fmarek-kindred, line 6: Are there any reasons why did you methods instead of constants?

fmarek-kindred (approved):

packages/talos_messenger_actions/src/kafka/models.rs
gk-kindred: I actually didn't like this approach as well, but serde supports assigning custom default values only via `functions` or `methods`. 

You can refer to this [long running issue](https://github.com/serde-rs/serde/issues/368) which was opened in 2016, and still active discussions on it.  

I looked at some suggestions, and thought maybe it's an overkill for now, and just kept it simple with the function

PR #90, opened 2023-10-18
feat: Change workflow trigger to manually build SDK

fmarek-kindred:


.github/workflows/npm_publish_light.yml
```diff
@@ -8,8 +8,8 @@
 
 name: Build and Publish Cohort SDK NPMs
 on:
-    #workflow_call:
-    push:
+    workflow_call:
```
gk-kindred, line 11: I think you should use [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) for manual trigger, instead of `workflow_call`

gk-kindred (approved):

PR #91, opened 2023-10-23
feat: Implement uninstall step in the workflow

fmarek-kindred:


gk-kindred:
The solution looks fine to unpublish.

Although I am thinking, do we need to go to this path of
-  `publish` then find something doesn't work and then `unpublish`
- Or we do some tests (minus ARM architecture, due to openssl), and then `publish` only if the tests pass?

gk-kindred (approved):

PR #94, opened 2023-11-13
feat: enable cohort to pass on_commit messages for messenger

gk-kindred:
- As this introduces a change in the contract by adding an option `onCommit`, will need to publish new package. Probably safe to still push as minor as no one really consumes the JS packages yet.
- The topic for `onCommit` is hardcoded in the example apps, to keep it simple for example sake, thereby minimising refactor.
- Load/performance tested with the rust version and performance has been pretty much in range to earlier commit. 
   - Expect 4000 / actual between 3200 to 3500
   - 
image

### DIFF
--- a/cohort_banking_initiator_js/src/banking-app.ts
+++ b/cohort_banking_initiator_js/src/banking-app.ts
@@ -148,7 +148,6 @@ export class BankingApp {
                     }
                 }
 
-                // TODO: GK - Add on commit action
             },
             snapshot: state.snapshotVersion,
             timeoutMs: 0,
--- a/cohort_banking_initiator_js/src/banking-app.ts
+++ b/cohort_banking_initiator_js/src/banking-app.ts
@@ -21,7 +21,7 @@ export class BankingApp {
         private pond: Pond,
         private database: Pool,
         private queue: BroadcastChannel,
-        private onFinishListener: (appRef: BankingApp) => any) {}
+        private onFinishListener: (appRef: BankingApp) => any) { }
 
     async init() {
         this.initiator = await Initiator.init(sdkConfig)
@@ -133,6 +133,22 @@ export class BankingApp {
                 writeset: [tx.from, tx.to],
                 readvers: state.items.map(i => i.version),
                 statemaps: [{ "TRANSFER": tx }],
+                onCommit: {
+                    publish: {
+                        kafka: [
+                            {
+                                topic: "test.transfer.feedback.js",
+                                value: {
+                                    "from_account": tx.from,
+                                    "to_account": tx.to,
+                                    "amount": tx.amount
+                                }
+                            },
+                        ]
+                    }
+                }
+
+                // TODO: GK - Add on commit action
             },
             snapshot: state.snapshotVersion,
             timeoutMs: 0,
@@ -143,14 +159,16 @@ export class BankingApp {
         let cnn: PoolClient
         try {
             cnn = await this.database.connect()
-            const result = await cnn.query({ name: "get-state", text:
-                `SELECT
+            const result = await cnn.query({
+                name: "get-state", text:
+                    `SELECT
                     ba."number" as "id", ba."version" as "version", cs."version" AS snapshot_version
                 FROM
                     bank_accounts ba, cohort_snapshot cs
                 WHERE
                     ba."number" = $1 OR ba."number" = $2`,
-                values: [tx.from, tx.to] }
+                values: [tx.from, tx.to]
+            }
             )
 
             if (result.rowCount != 2) {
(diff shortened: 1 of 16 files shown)


packages/cohort_banking/src/app.rs
```diff
@@ -64,19 +65,35 @@ impl BankingApp {
...
         let statemap = vec![HashMap::from([(
             BusinessActionType::TRANSFER.to_string(),
             TransferRequest::new(request.from.clone(), request.to.clone(), request.amount).json(),
         )])];
 
+        let on_commit_value = json!({
+            "publish": {
+                "kafka": [
+                    {
+                        "topic": "test.transfer.feedback",
```
gk-kindred, line 79: Hardcoded topic as it's just the example app, and to minimise refactor.

cohort_banking_initiator_js/src/banking-app.ts
```diff
@@ -133,6 +133,22 @@ export class BankingApp {
                 writeset: [tx.from, tx.to],
                 readvers: state.items.map(i => i.version),
                 statemaps: [{ "TRANSFER": tx }],
+                onCommit: {
+                    publish: {
+                        kafka: [
+                            {
+                                topic: "test.transfer.feedback.js",
```
gk-kindred, line 140: Hardcoded topic as it's just the example app, and to minimise refactor.

cohort_banking_initiator_js/src/banking-app.ts
```diff
@@ -133,6 +133,22 @@ export class BankingApp {
...
+                                    "from_account": tx.from,
+                                    "to_account": tx.to,
+                                    "amount": tx.amount
+                                }
+                            },
+                        ]
+                    }
+                }
+
+                // TODO: GK - Add on commit action
```
fmarek-kindred, line 151: Is this supported? If we support action why dont we add them to interface too?

packages/cohort_banking/src/model/requests.rs
```diff
@@ -6,7 +6,8 @@ pub struct CandidateData {
     pub readset: Vec<String>,
     pub writeset: Vec<String>,
     pub statemap: Option<Vec<HashMap<String, Value>>>,
-    // The "snapshot" is intentionally messing here. We will compute it ourselves before feeding this data to Talos
+    // The "snapshot" is intentionally missing here. We will compute it ourselves before feeding this data to Talos
+    pub on_commit: Option<Value>,
```
fmarek-kindred, line 10: Please add this to readme as well;

https://github.com/kindredgroup/talos/blob/master/cohort_sdk_client/README.md#about-jscertificationcandidate

cohort_banking_initiator_js/src/banking-app.ts
gk-kindred: This is a good point, messenger is written in a way that it can accept any valid json, but will only process the supported types. 

But on the other end, it would be nice if the sender is made aware on what they can send. Let me think on it and see how I can add this in at the cohort end. Thank for pointing this out!

packages/cohort_sdk_js/src/initiator/mod.rs
```diff
@@ -67,12 +67,73 @@ impl From<JsInitiatorConfig> for Config {
     }
 }
 
+// #[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+// #[serde(rename_all = "camelCase")]
+#[napi(object)]
+pub struct JSKafkaAction {
```
fmarek-kindred, line 73: Minor, not a blocker for PR merge. Just most of the data structures use "Js" prefix (here we seems to use "JS")

cohort_sdk_client/README.md
```diff
@@ -113,10 +138,11 @@ export interface JsCertificationCandidate {
 Before SDK can issue a certification request to Talos Certifier it needs some details from you. You will have to query your local database to fetch the following:
 1. Identifiers and version numbers of all objects involved in your transaction. These are known as `readset`, `writeset` and `readvers`.
 2. The copy of your transaction as one serializable object. It makes sense to describe your transaction as JSON object and serialise it to string. This is known as `statemap`.
+3. Any additional message to be published for candidate requests with committed decision outcome, can added to `onCommit` field. Currently the SDK supports only publishing to **Kafka**.
```
fmarek-kindred, line 141: Did we want to say "can be added to" ?

fmarek-kindred (approved):

packages/cohort_sdk_js/src/initiator/mod.rs
gk-kindred: Oops, that should be `Js` I will update that soon. Thanks for spotting it

PR #102, opened 2024-07-31
fix: messenger publish order and hang state 

gk-kindred:
- Fix for messenger not publishing messages in order
- Fix scenario when unknown message is received, catching and ignoring the error propogated by the consumer

### DIFF
--- a/packages/talos_messenger_core/src/services/inbound_service.rs
+++ b/packages/talos_messenger_core/src/services/inbound_service.rs
@@ -152,7 +152,7 @@ where
                         // 2.1 For CM - Install messages on the version
                         Ok(Some(ChannelMessage::Candidate(candidate))) => {
                             let version = candidate.message.version;
-                            info!("Candidate version received is {version}");
+                            debug!("Candidate version received is {version}");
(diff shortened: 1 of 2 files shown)


packages/talos_messenger_core/src/services/inbound_service.rs
```diff
@@ -145,14 +145,14 @@ where
...
 
-                    // 2. Add/update to suffix.
-                    match msg {
+                    match reciever_result {
                         // 2.1 For CM - Install messages on the version
-                        ChannelMessage::Candidate(candidate) => {
-
+                        Ok(Some(ChannelMessage::Candidate(candidate))) => {
                             let version = candidate.message.version;
+                            info!("Candidate version received is {version}");
```
fmarek-kindred, line 155: Is hits info important? Can it be debug instead of info?

gk-kindred: Aaah, it shouldn't be `info!` will update that to debug

fmarek-kindred (approved):

PR #107, opened 2024-09-09
feat: add missing decision headers and carry forward headers to on_commit actions

gk-kindred:


packages/talos_certifier/src/services/decision_outbox_service.rs
```diff
@@ -123,21 +112,35 @@ impl SystemService for DecisionOutboxService {
...
-                    Ok(decision) => {
-                        if let Err(publish_error) = DecisionOutboxService::publish_decision(&publisher, &decision, headers).await {
-                            error!(
-                                "Error publishing message for version={} with reason={:?}",
-                                decision.version,
-                                publish_error.to_string()
-                            );
+            tokio::spawn({
+                let decision_headers = DecisionHeaderBuilder::with_additional_headers(headers.into()).add_meta_headers(DecisionMetaHeaders::new(
+                    1_u64, // major version of decision message
```
fmarek-kindred, line 117: You might want to keep it somewhere like a constant or "static field", visible. Here it is a bit buried in the code logic.

fmarek-kindred (approved):

packages/talos_certifier/src/services/decision_outbox_service.rs
gk-kindred: Thanks for spotting that. Will update it today

PR #108, opened 2024-11-01
feat: messenger improve latency

gk-kindred:


packages/talos_messenger_actions/src/messenger_with_kafka.rs
```diff
@@ -93,6 +93,10 @@ pub struct Configuration {
     pub allowed_actions: HashMap<String, Vec<String>>,
     /// Channel buffer size for the internal channels between threads
     pub channel_buffers: Option<ChannelBuffers>,
+    /// Commit size to decide how often the certifier topic can be committed by the consumer.
+    /// The more ofter the commit is done has inverse impact on the latency.
```
fmarek-kindred, line 97: A small typo

packages/talos_messenger_core/src/services/inbound_service.rs
```diff
@@ -150,8 +196,28 @@ where
...
+                // Receive feedback from publisher.
+                Some(feedback_result) = self.rx_feedback_channel.recv() => {
+                    match feedback_result {
+                        MessengerChannelFeedback::Error(version, key, message_error) => {
+                            error!("Failed to process version={version} with error={message_error:?}");
+                            self.handle_action_failed(version, &key);
+
+                        },
+                        MessengerChannelFeedback::Success(version, key) => {
+                            info!("Successfully processed version={version} with action_key={key}");
```
fmarek-kindred, line 212: Should it be debug instead of info, to reduce a noise in the log?

fmarek-kindred (approved):
Given that suffix logic was updated, I wonder if we need to update unit tests as well.

PR #116, opened 2025-02-11
Expose more agent's metrics

fmarek-kindred:


gk-kindred (approved):

packages/talos_agent/src/messaging/api.rs
```diff
@@ -32,11 +32,20 @@ pub struct CandidateMessage {
     pub statemap: Option<StateMap>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub on_commit: Option<Box<Value>>,
+    pub certification_started_at: i128,
```
gk-kindred, line 35: Could you add comment like this maybe so it helps to easily remember when we revisit these fields in future?
```
/// cohort start certification
```

packages/talos_agent/src/messaging/api.rs
```diff
@@ -32,11 +32,20 @@ pub struct CandidateMessage {
...
+    pub request_created_at: i128,
```
gk-kindred, line 36:
```
/// talos agent request created for certifier
```

PR #118, opened 2025-02-12
Deserialise new metrics fields into default 0

fmarek-kindred:


gk-kindred (approved):
Nice 👍 on adding the tests

PR #119, opened 2025-02-19
Suffix prune and insert optimisations

gk-kindred:
replicator prune strategy has been optimised

inserting items into suffix has also been revised

### DIFF
--- a/packages/talos_cohort_replicator/src/core.rs
+++ b/packages/talos_cohort_replicator/src/core.rs
@@ -1,4 +1,4 @@
-use log::warn;
+use log::{info, warn};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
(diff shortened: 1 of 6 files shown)


packages/talos_cohort_replicator/src/core.rs
```diff
@@ -130,17 +130,28 @@ where
...
             }
+        } else {
+            self.next_commit_offset = Some(version)
         }
+        // }
     }
 
     pub(crate) async fn commit(&mut self) {
         if let Some(version) = self.next_commit_offset {
+            info!("Commit till offset {:?} ", version);
```
fmarek-kindred, line 154: Is "info" level intentional here?

fmarek-kindred (approved):

PR #126, opened 2025-03-24
Cohort replicator metrics.

fmarek-kindred:


packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -121,12 +173,17 @@ where
...
 
                     // Update the status flag
                     statemap_installer_queue.update_queue_item_state(&key, StatemapInstallState::Inflight);
                 }
+
+                metrics.record_sizes(installation_tx.capacity(), statemap_installer_queue.queue.len());
             }
             _ = cleanup_interval.tick() => {
+                metrics.record_sizes(installation_tx.capacity(), statemap_installer_queue.queue.len());
+
```
gk-kindred, line 186: Would be better to record this after the clean up method is called towards the end of the arm, as this will indirectly help in identifying the clean up happened and the size shrunk. 

```rust
statemap_installer_queue.remove_installed() 
```

gk-kindred (approved):

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
fmarek-kindred: I will move it down. However we would export those on the next tick anyway.

PR #128, opened 2025-04-02
feat: base for back pressure in talos ecosystem and usage in messenger

gk-kindred:
- Base struct `TalosBackPressureConfig` which could be added to any service in talos ecosystem with all the essential methods to capture and determine if back pressure is required.
- Usage of `TalosBackPressureConfig` in Talos messenger service.

### DIFF
--- a/packages/talos_common_utils/src/back_pressure.rs
+++ b/packages/talos_common_utils/src/back_pressure.rs
@@ -74,3 +74,157 @@ impl TalosBackPressureConfig {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::TalosBackPressureConfig;
+
+    #[test]
+    #[should_panic(expected = "min_threshold (Some(50)) must be less or equal to the max_threshold (Some(10))")]
+    fn test_panic_min_greater_than_max_threshold() {
+        let _ = TalosBackPressureConfig::new(Some(50), Some(10));
+    }
+
+    #[test]
+    fn test_back_pressure_when_min_equal_max_threshold() {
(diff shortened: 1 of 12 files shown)


packages/talos_common_utils/src/back_pressure.rs
```diff
@@ -0,0 +1,230 @@
+#[derive(Debug, Default, Clone)]
+pub struct TalosBackPressureConfig {
+    /// Current count which is used to check against max and min to determine if back-pressure should be applied.
+    pub current: u32,
+    /// Flag denotes if back pressure is enabled, and therefore the thread cannot receive more messages (candidate/decisions) to process.
+    pub is_enabled: bool,
+    /// Max count before back pressue is enabled.
+    /// if `None`, back pressure logic will not apply.
+    pub max_threshold: Option<u32>,
```
fmarek-kindred, line 9: If this is "max count" then its better if we named the attribute with workd "count" in it. 

Please change the order. 
`is enabled`
then`count` and `max count`.

packages/talos_common_utils/src/back_pressure.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// `min_threshold` helps to prevent immediate toggle between switch on and off of the backpressure.
+    /// Batch of items to process, when back pressure is enabled before disable logic is checked?
+    /// if None, no minimum check is done, and as soon as the count is below the max_threshold, back pressure is disabled.
```
fmarek-kindred, line 12: Please have another look at the description. I am not able to follow the purpose of this parameter.

PR #130, opened 2025-04-16
chore: certifier suffix reset and commit with minimum offset threshold

gk-kindred:
This change covers the work needed to maintain a minimum offset from committing so that when the certifier restarts or when the suffix is reset, the suffix will be hydrated with sufficient candidates to avoid initial aborts

### DIFF
--- a/packages/talos_certifier_adapters/Cargo.toml
+++ b/packages/talos_certifier_adapters/Cargo.toml
@@ -22,9 +22,11 @@ env_logger = { workspace = true }
 # Async
 tokio = { workspace = true }
 async-trait = { workspace = true }
-futures-util = "0.3.21"
+# Futures
+futures-util = "0.3.26"
+futures-executor = "0.3.28"
 # Kafka
-rdkafka = { version = "0.34.0", features = ["sasl"]  }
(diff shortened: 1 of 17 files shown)


packages/talos_certifier_adapters/Cargo.toml
```diff
@@ -48,10 +50,10 @@ thiserror = "1.0.31"
 mockall = "0.11.0"
 
 # internal crates
-talos_metrics =       { path = "../talos_metrics", version = "0.2.41-dev" }
-talos_certifier =     { path = "../talos_certifier", version = "0.2.41-dev" }
-talos_suffix =        { path = "../talos_suffix", version = "0.2.41-dev" }
-talos_common_utils =  { path = "../talos_common_utils", version = "0.2.41-dev" }
+talos_metrics = { path = "../talos_metrics", version = "0.2.41-dev" }
```
fmarek-kindred, line 53: This is a bit old. 42 is released.

gk-kindred: Thanks, the PR also pointed out this conflict. I will rebase against master

fmarek-kindred (approved):

PR #136, opened 2025-05-07
feat: fix replicator not send for install

gk-kindred:
### Major changes as part of this PR are:-

- Update the `snapshot_version` on rebalance.
- Add new metrics gauge to capture `snapshot_version`
- Exhaustive tests around `statemap_queue_service`
- Changed the logic on how feedbacks are use for pruning logic.


packages/talos_certifier_adapters/src/kafka/contexts/replicator_consumer_context.rs
```diff
@@ -0,0 +1,48 @@
...
+    T: ResetVariantTrait + Clone + Send + Sync,
+{
+    fn post_rebalance(&self, rebalance: &rdkafka::consumer::Rebalance<'_>) {
+        match rebalance {
+            Rebalance::Assign(partitions) => {
+                info!("Rebalance complete and partition of certification topic: {:?} is assigned", self.topic);
+                let tpl_vec = partitions.elements_for_topic(&self.topic);
+                if !tpl_vec.is_empty() {
+                    if let Err(err) = block_on(try_send_with_retry(
+                        &self.message_channel_tx,
```
fmarek-kindred, line 36: Is it safe to mix this with tokio?

packages/talos_common_utils/src/sync/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+
+                // If reached max attempts, break.
+                if let Some(max_attempts) = config.max_retry_attemptes {
+                    if counter >= max_attempts {
+                        break Err(TrySendError::Full(message));
+                    }
+                }
+                // If reached max allowed duration, break.
+                if let Some(max_duration_ms) = config.max_retry_duration_ms {
+                    if start.elapsed().as_millis() + sleep_time as u128 >= max_duration_ms as u128 {
```
fmarek-kindred, line 55: To me it feels like if you you would precompute anticipated abort time (now + config.max_retry_duration_ms) and use it instead of `start` then logic would be simpler? Would be no need to do any more calculations, just measure diff between now and anticipated abort time

packages/talos_common_utils/src/lib.rs
```diff
@@ -1,2 +1,13 @@
 pub mod env;
 pub mod otel;
+pub mod sync;
+
+pub trait ResetVariantTrait {
```
fmarek-kindred, line 5: Do you think the re-use of suffix and all those so many traits actually simplify code? I am wondering if having 3 distinct suffixes and no traits at all would make maintaining the code easier.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -84,146 +97,269 @@ impl Metrics {
...
+
+            self.metrics.record_snapshot(snapshot_version);
+            Some(snapshot_version)
+        } else {
+            None
+        }
+    }
+
+    /// If a version is not on the queue, get the nearest version below the provided version from queue.
+    fn get_nearest_valid_version(&self, version: u64) -> Option<u64> {
```
fmarek-kindred, line 166: this method uses 3 loops. Can we use one single classic loop via indices to compute what we need?

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -84,146 +97,269 @@ impl Metrics {
...
+                            version,
+                            safepoint,
+                            state
+                        });
+
+
+                    },
+                    // Update the snapshot value.
+                    Some(StatemapQueueChannelMessage::UpdateSnapshot) => {
+                        info!("Fetching latest snapshot version from callback");
```
fmarek-kindred, line 245: Lets remove this one or add some information to it. Currently it tells us nothing useful from troubleshooting point of view. Just some string in the log. And there will be lots of them.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -84,146 +97,269 @@ impl Metrics {
...
+                        // Update the snapshot with the latest from callback only if it is greater than our internal snapshot tracker.
+                        if self.update_snapshot(snapshot_version_from_callback).is_some() {
+                            // The snapshot version we updated may not be present in the queue, so we get the nearest one below this snapshot version and prune till that version
+                            // If there are no versions below this one, then there is no need to prune.
+                            if let Some(version) = self.get_nearest_valid_version(snapshot_version_from_callback) {
+                                // prune items till the specified version.
+                                self.statemap_queue.prune_till_version(version);
+                                // Inform replicator service to remove all versions below this index.
+                                if let Err(err) = try_send_with_retry(&self.replicator_feedback, ReplicatorChannel::LastInstalledVersion(version), TrySendWithRetryConfig::default()).await {
+                                    error!("Failed to send LastInstalledVersion {version} with error {err:?}");
```
fmarek-kindred, line 257: "Connect" this log statement with snapshot value, so that we know the full context.

packages/talos_certifier_adapters/src/kafka/contexts/replicator_consumer_context.rs
gk-kindred: This is a good point and thanks for spotting this. It was a lazy way to implement async into the sync code block. Although it works, it is not ideal as it could block the current thread. I will update it use `tokio::spawn`

packages/talos_common_utils/src/lib.rs
gk-kindred: This is something I have thought a couple of times, as the suffix was original created to cater to `certifier` and then it got carried over to `replicator` and `messenger` as they are similar in many ways and different in some other ways.  Over time, the are diverging more.

Probably worth looking at a later time on having three separate suffices and what is common can be shared via trait. I still do believe a common trait adds value here, but probably not in its current form.

It is too much to take up at the moment and we can keep a ticket to revisit on this.

packages/talos_certifier_adapters/src/kafka/contexts/replicator_consumer_context.rs
```diff
@@ -0,0 +1,48 @@
...
+    consumer::{ConsumerContext, Rebalance},
+    ClientContext,
+};
+use talos_common_utils::{
+    sync::{try_send_with_retry, TrySendWithRetryConfig},
+    ResetVariantTrait,
+};
+
+/// Replicators's consumer context used to detect a rebalance and update the snapshot.
+pub struct ReplicatorConsumerContext<T>
```
fmarek-kindred, line 16: It is better and much simpler if this structure be moved to `talos_cohort_replicator` package. Its implementation of `ConsumerContext` is specific to replicator scenario and different from certifier context. There is no reason for this to be in the shared library. 

I am still figuring out what is ResetVariantTrait, why it is needed, instead of us simply sending a specific message via channel.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -84,146 +97,269 @@ impl Metrics {
...
+
+
+                    },
+                    // Update the snapshot value.
+                    Some(StatemapQueueChannelMessage::UpdateSnapshot) => {
+                        info!("Fetching latest snapshot version from callback");
+                        let snapshot_version_from_callback = self.get_latest_snapshot_from_callback().await?;
+
+                        // Update the snapshot with the latest from callback only if it is greater than our internal snapshot tracker.
+                        if self.update_snapshot(snapshot_version_from_callback).is_some() {
```
fmarek-kindred, line 249: Here we are handling a feedback of successful install. Can you please provide a comment describing in what scenario the update snapshot would return None? If it is somehow possible that our snapshot stored internally in the queue has jumped ahead, we better have this case documented.

packages/talos_common_utils/src/sync/mod.rs
gk-kindred: Agree, I will update this.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
fmarek-kindred: No, not here, its is below. I am removing this comment.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -84,146 +97,269 @@ impl Metrics {
...
-                        let enc_time = statemap_installer_queue.update_queue_item_state(&key, StatemapInstallState::Installed);
-                        metrics.record_latency(enc_time.map(|enqueue_time_nanos| Duration::from_nanos((OffsetDateTime::now_utc().unix_timestamp_nanos() - enqueue_time_nanos) as u64)));
-
-                        if let Some(last_contiguous_install_item) = statemap_installer_queue.queue.iter().take_while(|(_, statemap_installer_item)| statemap_installer_item.state == StatemapInstallState::Installed).last(){
-                            statemap_installer_queue.update_snapshot(last_contiguous_install_item.1.version) ;
+                        let enc_time = self.statemap_queue.update_queue_item_state(&version, StatemapInstallState::Installed);
+                        self.metrics.record_latency(enc_time.map(|enqueue_time_nanos| Duration::from_nanos((OffsetDateTime::now_utc().unix_timestamp_nanos() - enqueue_time_nanos) as u64)));
+
+                        if let Some(version) = self.statemap_queue.get_last_contiguous_installed_version(){
+                            if self.update_snapshot(version).is_some() {
```
fmarek-kindred, line 284: Wanted to comment on that one :)

Here we are handling a feedback of successful install. Can you please provide a comment describing in what scenario the update snapshot would return None? If it is somehow possible that our snapshot stored internally in the queue has jumped ahead, we better have this case documented.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
gk-kindred: Can you elaborate on this comment? I didn't understand.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -84,146 +97,269 @@ impl Metrics {
...
-                                        | success={installation_success_count}
-                                        | awaiting_installs={awaiting_count}
-                                        | inflight_count={inflight_count}
-                        current snapshot: {}
-                        \n ", statemap_installer_queue.snapshot_version);
-                    //   last vers send to install : {last_item_send_for_install}
-
-                    if config.enable_stats && awaiting_count > 0 && inflight_count == 0
+
+                if self.config.enable_stats {
```
fmarek-kindred, line 330: Can we please have a dedicated method/function to deal with this? This method is already big.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -84,146 +97,269 @@ impl Metrics {
...
+                    let duration_sec = Duration::from_nanos((self.metrics.last_install_end - self.metrics.first_install_start) as u64).as_secs_f32();
+                    let tps = self.metrics.installation_success_count as f32 / duration_sec;
```
fmarek-kindred, line 332: Please make a dedicated method in metrics object to encapsulate tps computation. There is no need for installation_success_count and other fields to be publicly mutable.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
gk-kindred: True, it started as a debug for me to know which instance is fetching new version after rebalance. I will remove that and add an info below the callback function with the new version received. As that would be more helpful.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
gk-kindred: `LastInstalledVersion` is actually the snapshot version. I will update the log to make that obvious.

```
Failed to send latest LastInstalledVersion..
```

to 
```
Failed to send latest snapshot_version..
```

packages/talos_certifier_adapters/src/kafka/contexts/replicator_consumer_context.rs
gk-kindred: The reason why it is not put into `talos_cohort_replicator` is on purpose. This is because, while designing `certifier` and then later `replicator` and `messenger`,  **_port - adapter_** architecture was followed, were the core part of these services are isolated for the abcast or db. 

Ideally `talos_certifier_adapters` should be renamed to  a more common name, or replicator have its own adapter package. 

But the only thing extra from `talos_certifier_adapters` that replicator needs is the specific consumer context, that is why I didn't want to create a new package. 

Also I didn't want to add it to `talos_cohort_replicator`, because I will then need to bring in `kafka` related dependencies into that package, which I don't prefer.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
gk-kindred: I am summarising this as a comment in the code to help understand this better

packages/talos_common_utils/src/lib.rs
fmarek-kindred: I prefer us to find a middle ground. Sure not to split suffixes now, but at least have a look if we can avoid having this new trait. Why can't specific kafka context send a specific kafka message, why does it need to have interface?

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
gk-kindred: I have updated the code, to remove some part of it  + move the rest into a new method

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
fmarek-kindred: `self.statemap_queue.queue.get_index_of` - is a lookup 1 through the list which will have a loop inside it.
`self.statemap_queue.queue.iter().take_while` - is a lookup 2 through the list
`self.statemap_queue.queue.get_index_of`  - is a lookup 3 through the same list.

can we have something like
```

version_index = -1
last_below = (-1, -1) // index, version
for i = array.start till array..end {
  if array[ index ] == version then version_index = i // first case is solved
  if array[ index ] < version && array[ index ] > last_below.0  then last_below = (index, array [index ] ) // 2nd and 3rd cases solved
}

if version_index == -1 return

if ... then return last_below.1

return none
```

packages/talos_common_utils/src/lib.rs
gk-kindred: It is to do with how the packages are currently. 

`ReplicatorConsumerContext` is in `talos_certifier_adaptors` package.  - This is because I don't want to bring in `kafka` related dependencies into `talos_cohort_replicator` package.

`StatemapQueueChannelMessage::UpdateSnapshot` which is the message we need to publish via the `ReplicatorConsumerContext` is in `talos_cohort_replicator` package which I didn't want to bring into `talos_certifier_adaptors` as dependency. 

So in this scenario using the above trait helps avoid the circular dependency.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
gk-kindred: `get_index_of` is from `IndexMap` with O(1) computation. It is not expensive. There is only 1 loop `self.statemap_queue.queue.iter().take_while`.  I don't prefer the code snippet above, as this shouldn't be expensive.

packages/talos_certifier_adapters/src/kafka/contexts/replicator_consumer_context.rs
fmarek-kindred: Why you dont not want to bring kafka into replicator? Can replicator work without kafka? :) It uses only kafka and will always use only kafka. This adaptor is making things more complicated than they should be. Replicator core is already well incapsulated behind channels. All signals come through channel messages. According to framework design you start abstracting things into a common interface when you anticipate multiple implementations. I dont think the second messaging platform will ever be incorporated here. Lets be realistic.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
fmarek-kindred: Yes, thanks. I understood about case with rebalance. But with feedbacks to come out of order is something to think about (at least for me).

packages/talos_certifier_adapters/src/kafka/contexts/replicator_consumer_context.rs
gk-kindred: The idea during the initial days while creating it as open source was, all these services could support any abcast or  database. And that is why the core part of `certifier`, `replicator` or `messenger` is isolated from what is putting the data into it, or where it is pushing the data to. It is all hidden behind traits. 

Although it is true internally we support only kafka, and that is why there are certain packages which isolate it. They are written in a way that it could be easily switched to another abcast

packages/talos_common_utils/src/sync/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// - Defaults to 30ms.
+    pub max_sleep_ms: u32,
+    /// Overall max duration to attempt retries.
+    /// - Use `None` to disable this check.
+    /// - Defaults to 1 seconds.
+    pub max_retry_duration_ms: Option<u32>,
+    /// Overall number of retries.
+    /// - Use `None` to disable this check.
+    /// - Defaults to 5 retries.
+    pub max_retry_attemptes: Option<u32>,
```
fmarek-kindred, line 18: typo in `attempts`

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -29,44 +30,87 @@ impl Default for StatemapQueueServiceConfig {
...
+            ..Self::default()
+        };
         if let Some(meter) = meter {
             Self {
                 enabled: true,
                 h_installation_latency: Some(meter.f64_histogram("repl_statemap_queue_latency").build()),
                 g_installation_tx_channel_usage: Some(meter.u64_gauge("repl_install_channel").with_unit("items").build()),
                 g_statemap_queue_length: Some(meter.u64_gauge("repl_statemap_queue").with_unit("items").build()),
+                g_snapshot_version: Some(meter.u64_gauge("repl_statemap_queue_snapshot").with_unit("items").build()),
+                g_installation_tps: Some(meter.f64_gauge("repl_statemap_queue_installation_tps").with_unit("items").build()),
```
fmarek-kindred, line 85: In general regarding naming:
OTEL naming recommendation suggests that we avoid putting unit into metric name. For example if we are counting received items then metric name should be `received`, unit `items` and meter `count`. Then when OTEL will be sending this metric value to backend it will concatenate them appropriately.  
When we do not know or do not need unit we can skip it. 
What we want to avoid is having unit be a part of metric name (tps) and also added via API  (items). This will produce weird looking metric inside backend, something like `repl_statemap_queue_installation_tps_items`

Specifically regarding `repl_statemap_queue_installation`. Our statemap queue is not installing much, right? Installation tps would. be property of installer service. The metric you are recording looks more like the rate at which you receive the feedback messages from the feedback channel.
As these messages can come with Ok or Err flavour, I suggest to record them as simple count with generic `is_success` label. Ok means is_success = true, and Err means false. Then when doing UI for this metric we can produce rate on the graph because we will have count and time for each measurement. `rate()` function in Prometheus/ Grafana is designed for that use case. Additional label `is_success` will allow us to plot three lines if we wanted - combined (ignoring label), plus is_sucess=true and is_sucess=false. Metric with label `is_success=false` can be also added to generic errors panel to show as gauge.

```
match install_result {
   OK => metric.feedback_received(true)
   Err => metric.feedback_received(false)
}
```

For inner debug stats you may compute tps as you do already but that would be hidden inside `feedback_received` method and printed to log when needed. Sending it to OTEL would not be necessary. 

The next part is not applicable to replicator but for the context just sharing with you why OTEL is discouraging from sending aggregated metrics is that devs usually forget about HA / multipod deployment. Sometimes when developer is trying to compute TPS of HTTP method they are actually recording only TPS of one pod, which will screw up the full picture as every pod will have different TPS value. But sending only counter to OTEL will allow TPS graph to be computed by aggregating these counts from all pods and showing proper system throughput.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
fmarek-kindred: If it is O1 then yes 2 loops are removed.

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
```diff
@@ -84,183 +128,277 @@ impl Metrics {
...
+                        break;
+                    }
+                }
+            }
+            // Update the status flag
+        }
+    }
+    /// Internal logs and stats to help while debugging.
+    fn print_stats(&self) {
+        let tps = self.metrics.get_installation_tracker().get_tps();
```
fmarek-kindred, line 237: this is nice and easy

fmarek-kindred (approved):

packages/talos_cohort_replicator/src/services/statemap_queue_service.rs
gk-kindred: Thanks, I completely overlooked multiple instance of pods running, also didn't know the recommended practice from otel. I have removed this `Gauge` as I can the tps from the `snapshot_version` updates, which is on another Gauge.

packages/talos_certifier_adapters/src/kafka/contexts/replicator_consumer_context.rs
ekoutanov: A reason to abstract an abcast implementation may be for mocking/testing.

